PR #20, opened 2024-11-30
Issue 8 vuetify colour theme config

### LINKED ISSUE #8: Vuetify Colour Theme Config
Once the theme colours have been decided in the figma wireframes (https://github.com/codersforcauses/bluecrew/issues/2), the theme colours should be set in the ``main.ts`` file so that we have an easy way to ensure that the developers use consistent colours. See https://vuetifyjs.com/en/features/theme/

yuwengchoong:
## Change Summary
Added vuetify colour themes using the hexcodes from the figma team

- [yes] The pull request title has an issue number
- [ ] The change has tests
- [yes] The change has documentation

## Other Information

# Related issue

- Resolve #8

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -18,11 +18,11 @@ const app = createApp(App)
 // Object to store colour theme hexcodes
 const colourTheme = {
   colors: {
-    primary: '#193855',
-    titles: '#D12974',
-    accent: '#3FBEE0',
-    containers: '#E9DAC4',
-    text: '#4F4F4F',
+    primaryBlue: '#193855',
+    pink: '#D12974',
+    lightBlue: '#3FBEE0',
+    creamWhite: '#E9DAC4',
+    textGrey: '#4F4F4F',
   },
 }
 
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -15,11 +15,23 @@ import router from './router'
 
 const app = createApp(App)
 
+const colourTheme = {
+  colors: {
+    primary: '#193855',
+    secondary: '#D12974',
+    accent: '#3FBEE0',
+    containers: '#E9DAC4'
+  }
+}
+
 const vuetify = createVuetify({
   components,
   directives,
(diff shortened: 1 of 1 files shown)


client/src/main.ts
```diff
@@ -15,11 +15,25 @@ import router from './router'
 
 const app = createApp(App)
 
+// Object to store colour theme hexcodes
+const colourTheme = {
+  colors: {
+    primary: '#193855',
+    titles: '#D12974',
```
HuxleyBerry: Could you give these colours names which give more of an indication of the colour? For example, could the ``titles`` colour be renamed to something like ``primaryPink``?

HuxleyBerry (changes requested):

client/src/main.ts
yuwengchoong: Yup, I've changed the colour names accordingly

client/src/main.ts
```diff
@@ -19,7 +19,7 @@ const app = createApp(App)
 const colourTheme = {
   colors: {
     primaryBlue: '#193855',
-    pink: '#D12974',
+    primaryPink: '#D12974',
     lightBlue: '#3FBEE0',
     creamWhite: '#E9DAC4',
     textGrey: '#4F4F4F',
```
HuxleyBerry: Sorry to keep requesting trivial changes, but could you change the name of ``textGrey``? The reason is that when we use this colour in the components, we will be doing something like ``<p class="text-textGrey">Hello</p>``. IMO it looks a little dumb to have "text" twice in a row like that. So can you change "text" to some other word? Some options could be "theme", "custom", or "primary".

HuxleyBerry (changes requested):

client/src/main.ts
yuwengchoong: All good. I've changed it

HuxleyBerry (approved):
Nice 👍, lgtm

PR #30, opened 2024-12-07
Issue 16 create challenge db model

### LINKED ISSUE #16: Create Challenge DB Model
Create a django model for challenges. The fields should be:
- id (primary key)
- name
- description (can be blank)
- challenge type: one of "connect", "understand", or "act"
- points -- the amount of points users are to be awarded upon completing the challenge
- total completions -- we have this field so that the bluecrew admins have an easy way of quantifying the impact that the website is having.

Django model docs: https://docs.djangoproject.com/en/5.1/topics/db/models/

bibonut:
## Change Summary
Created Challenge DB model

### Change Form

- [X] The change has tests - django unit tests
- [X] The change has documentation - comments

## Other Information
N/A

# Related issue

- Resolve #16


server/bingo/tests.py
```diff
@@ -1,3 +1,21 @@
...
+            description="This is a test challenge.",
+            challenge_type="act",
+            points=5,
+            total_completions=10,
+        )
+
+    def test_create_challenge(self):
+        challenge = Challenge.objects.get(name="Test")
+        self.assertEqual(challenge.description, "This is a test challenge.")
+        self.assertEqual(challenge.challenge_type, "act")
```
HuxleyBerry, line 19: Can you add a test which checks that there is an error if we try to create a challenge with ``challenge_type`` something other than the allowed values?

server/bingo/tests.py
```diff
@@ -1,3 +1,21 @@
...
+        self.assertEqual(challenge.points, 5)
+        self.assertEqual(challenge.total_completions, 10)
+        self.assertEqual(str(challenge), "Challenge 1: Test (Act)")
```
HuxleyBerry, line 22: Maybe add a test which checks that ``total_completions`` is indeed set to the default value of ``0`` if it isn't specified.

HuxleyBerry (changes requested):
LGTM, I just have a couple of minor suggestions for additional things to test.

HuxleyBerry (commented):
Looks good, thanks. I'm gonna wait to approve this however since I'd like to merge the custom user model in first since its migrations need to be applied first. Hopefully it gets finished soon.

HuxleyBerry (approved):
LGTM

PR #32, opened 2024-12-07
Issue 3 create custom user model

### LINKED ISSUE #3: Create Custom User Model
The fields required (possibly subject to change) are:
- user_id (primary key)
- username
- First Name
- Last Name
- Email
- Password (should be stored as hash)
- Bio
- Total points (stored as an integer)
- Visibility preference (public/friends/bluecrew)
- Avatar (stored as an integer)

See these docs for more info: https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#specifying-a-custom-user-model and https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#a-full-example

SafetyInObscurity:
## Change Summary
Created basic user models

### Change Form

- [ ] The change has tests

# Related issue

- Resolve #3


server/bingo/models.py
```diff
@@ -1,3 +1,77 @@
...
+        )
+        user.is_admin = True
+        user.save(using=self.db)
+        return user
+
+
+class User(AbstractBaseUser):
+    REQUIRED_FIELDS = ["email"]
+
+    user_id = models.BigAutoField(primary_key=True)
```
HuxleyBerry: I think a ``BigAutoField`` is unnecessary here. You can change it to an ``AutoField``.

server/bingo/models.py
```diff
@@ -1,3 +1,77 @@
...
+    class Visibility(models.IntegerChoices):
+        PUBLIC = (2, "Public")
+        FRIENDS = (1, "Friends Only")
+        BLUECREW = (0, "BlueCrew Only")
+    visibility = models.IntegerField(
+        choices=Visibility,
+        blank=False,
+        default=0
+    )
+    avatar = models.IntegerField(default=0)
```
HuxleyBerry: Can you modify this field to only accept numbers from 0 to 5? The user will be given six options for their choice of avatar.

server/bingo/models.py
```diff
@@ -1,3 +1,77 @@
...
+
+    is_active = models.BooleanField(default=True)
+    is_admin = models.BooleanField(default=False)
+
+    objects = UserManager()
+
+    def has_perm(self, perm, obj=None):
```
HuxleyBerry: If you inherit from ``PermissionsMixin``, you won't need to include the following three methods.

HuxleyBerry (changes requested):
Looks good! I've left a few comments.

Additionally, I'm thinking it could be a good idea to have a different app for each "aspect of the backend" to keep things organised. How do you feel about renaming the app that your code is in from "bingo" to "users" or something like that? Then later on, we'd make a new app for challenges, ect. Also, feel free to rename the "api" folder to "config" as you suggested earlier.

server/bingo/models.py
```diff
@@ -1,3 +1,77 @@
...
+            username=username,
+            email=email,
+            password=password
+        )
+        user.is_admin = True
+        user.save(using=self.db)
+        return user
+
+
+class User(AbstractBaseUser):
```
HuxleyBerry: Oops, I forgot to save this in my above review, but I had also asked you to inherit from ``PermissionsMixin`` here: https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#custom-users-and-permissions

HuxleyBerry (changes requested):

server/bingo/models.py
```diff
@@ -1,3 +1,77 @@
-from django.db import models # noqa
+from django.db import models
+from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
 
-# Create your models here.
+
+class UserManager(BaseUserManager):
+
+    def create_user(self, username, email, password=None):
```
HuxleyBerry: This method (and the ``create_superuser`` method) should also take in keyword arguments for the other fields on the model. This will allow creating user instances with the necessary fields filled out through this method. So the method signature should be something like ``create_user(self, username, email, password, **extra_fields)``.

Was there any reason for having the default value of ``password`` being ``None`` btw? If not, I think that default value should be removed.

HuxleyBerry (changes requested):

HuxleyBerry:
There are a couple of other things that could reasonably be put under the scope of this issue, namely adding some additional fields (gender identity and indigenous status per a recent update from the client) and writing some tests for this model. But I think we want this model in main ASAP, so I think we can leave implementing those for a later issue.

SafetyInObscurity:
All requested changes have been made. Not sure creating an app just for users is appropriate considering that no other systems are currently intended to interact with these user accounts. Probably worth discussing with the client.

server/bingo/models.py
```diff
@@ -1,3 +1,73 @@
...
+
+class User(AbstractBaseUser, PermissionsMixin):
+    REQUIRED_FIELDS = ["email"]
+
+    user_id = models.AutoField(primary_key=True)
+
+    username = models.CharField(max_length=30, unique=True)
+    USERNAME_FIELD = "username"
+
+    first_name = models.CharField(max_length=30)
```
HuxleyBerry: ``first_name``, ``last_name``, and ``bio`` should have ``blank=True``.

HuxleyBerry (changes requested):
Sorry, one more thing which I should've noticed earlier but didn't.

server/bingo/models.py
```diff
@@ -1,3 +1,73 @@
...
+    last_name = models.CharField(max_length=30)
+    bio = models.CharField(max_length=300)
+    total_points = models.IntegerField(default=0)
+    birthdate = models.DateField(null=True)
```
HuxleyBerry: I think the ``birthdate`` field should have ``null=True`` removed and instead have some meaningless default. If you leave it as it is, you can't press save on the admin in the Django admin dashboard because a "This field is required" error pops up for the birthdate field. An alternative option is setting ``blank=True``, but I think the former option is better.

HuxleyBerry:
I'm going to make an exception and fix the requested changes myself since it's a high priority to get this issue over with and it's my fault for not noticing earlier the things I most recently requested changes for for.

HuxleyBerry (approved):

server/bingo/models.py
SafetyInObscurity: Done

server/bingo/models.py
SafetyInObscurity: I have enabled blank entries for emails, but I believe that in the end, blank birthdates shouldn't be allowed

SafetyInObscurity:
You beat me to it by 3 minutes!

HuxleyBerry:
> You beat me to it by 3 minutes!

Haha, turns out I could have left it to you then!

Issue #29, opened 2024-12-07
Create API view to get the details of the current user


HuxleyBerry:
## Prerequisites
#3 

## Issue
Create an API view which returns all the information the frontend needs to know about the currently logged in user. This view should be protected to only work for when the user is logged in.

PR #38, opened 2024-12-13
Issue 14 create user pinia store

### LINKED ISSUE #14: Create user pinia store
We want to have information about the currently logged-in user accessible from anywhere on the website. For this project, we will be using a [pinia store](https://pinia.vuejs.org/introduction.html) to do that.

The state of the store should be the details of the currently logged in user together with the access and refresh tokens. I suggest defining a new type to hold all the user details. Also, we want these details and the tokens to persist when the user closes the website. To achieve that, we can store them in the browser's local storage. We can use Vueuse's useStorage for this: https://vueuse.org/core/useStorage/.

Use the setup store syntax rather than the options store syntax (https://pinia.vuejs.org/core-concepts/#Setup-Stores).

mye9:
## Change Summary
Added a pinia store in ```client/src/stores/user.ts``` with the UserData (include userName, name and email), accessToken and refreshToken.

### Change Form

- [x] The change has tests - included in ```client/src/components/TestPiniaStore.vue```

## Other Information
N/A

# Related issue

- Resolve #14


client/src/components/TestPiniaStore.vue
HuxleyBerry, line 1: This is nice, but it should be removed before the branch gets merged into main. The PR template is a bit misleading -- you don't need to tick every box if it's not applicable. In particular, I don't expect tests for frontend issues.

client/src/stores/user.ts
```diff
@@ -0,0 +1,41 @@
+import { defineStore } from 'pinia'
+import { useStorage } from '@vueuse/core'
+import { computed } from 'vue'
+
+// User interface (include userName, name and email)
+interface User {
```
HuxleyBerry: To keep things organised, can you make a ``types`` directory in the ``src`` directory move this type definition into a file called ``user.ts`` within the ``types`` directory? Additionally, the ``User`` type should have several additional properties. It should have pretty much have a property for each of the fields in the [current backend user model](https://github.com/codersforcauses/bluecrew/blob/c2ccd82bd2535e4aaf0783dcddbe4ea8150a7aff/server/bingo/models.py). I think the required properties should be as follows:
- userId
- userName
- firstName
- lastName
- bio
- totalPoints
- email
- visibility
- avatar

Note that all these properties are in camelCase.

client/src/stores/user.ts
```diff
@@ -0,0 +1,41 @@
...
+  name: string
+  email: string
+}
+
+export const useUserStore = defineStore('user', () => {
+  const userData = useStorage<User | null>('userData', null)
+  const accessToken = useStorage<string | null>('accessToken', null)
+  const refreshToken = useStorage<string | null>('refreshToken', null)
+
+  const isLoggedIn = computed(() => !!userData.value)
```
HuxleyBerry: Can you change ``!!userData.value`` to ``userData.value !== null``? It's less confusing what's going on that way.

client/src/stores/user.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  // Login Status
+  const login = (user: User, tokens: { accessToken: string; refreshToken: string }) => {
```
HuxleyBerry: The ``login()`` function should be something a little different. It should take in login credentials (username and password), and then send a post request to ``api/token/`` to receive the tokens, and then store these tokens in the store. After that a get request should be sent to receive the details of the current user (this endpoint hasn't been implemented yet, that's [issue 29](https://github.com/codersforcauses/bluecrew/issues/29)) and that also gets stored in the store.

However, this isn't part of the scope of this issue, so for now you can just remove the function.

HuxleyBerry (changes requested):
Looks good! I have a few minor comments.

client/src/stores/user.ts
mye9: Sorry, I am just a little bit confused with the property of indigenousStatus. Does it refer to if the user is a superuser or not?

HuxleyBerry: It's to track whether the user identifies as indigenous (Aboriginal or Torres Strait Islander). But come to think of it, there's no need to be storing that in the frontend. The same applies with gender. I'll remove those two from the above list.

mye9: Thank you so much for the explanations. I will fix them as well.

mye9:
Hi Huxley, I would like to know if I should still keep the logout function besides the login since I already know the login should be achieved differently. If there are any other issues, please let me know so I can fix them. I really appreciate the help.

HuxleyBerry:
> Hi Huxley, I would like to know if I should still keep the logout function besides the login since I already know the login should be achieved differently. If there are any other issues, please let me know so I can fix them. I really appreciate the help.

Nope, your logout function does exactly what it needs to!

mye9:
Sorry, I just forgot to commit the changes within the ```stores/user.ts```, now everything is good to go

client/src/stores/user.ts
```diff
@@ -0,0 +1,28 @@
+import { defineStore } from 'pinia'
+import { useStorage } from '@vueuse/core'
+import { computed } from 'vue'
+import type { User } from '@/types/user'
+
+export const useUserStore = defineStore('user', () => {
+  //Import userData from types/user
```
HuxleyBerry: What does this comment mean?

mye9: So I just wanna say the next line is import from ```types/user.ts```, if it does not make sense I can just delete it.

HuxleyBerry: Ah I see. I'd be more accurate to say something like "the type below is imported from types/user", but that's not really necessary to say since you can just look at the import statement above.

mye9: Gotcha. I just deleted the sentence so as not to confuse it, and I believe everything else is good to go.

HuxleyBerry:
> Sorry, I just forgot to commit the changes within the `stores/user.ts`, now everything is good to go

Only thing left to do is remove the ``TestPiniaStore.vue`` file, then I'll approve!

mye9:
I just removed the ```TestPiniaStore.vue```, and I really appreciate the review. Thank you so much!

HuxleyBerry (approved):

PR #39, opened 2024-12-13
Issue 23 make footer component

### LINKED ISSUE #23: Make footer component
Make footer component as per the figma:
image

Then import and use the component in ``App.vue`` so that it appears on all the pages.

mye9:
## Change Summary
Added Footer component and fit for different versions of mobile and desktop

### Change Form

- [NA] The change has tests

## Other Information

# Related issue

- Resolve #23

### DIFF
--- a/client/src/App.vue
+++ b/client/src/App.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { RouterView } from 'vue-router'
-import BluecrewFooter from  "@/components/BluecrewFooter.vue";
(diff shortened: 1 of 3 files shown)


HuxleyBerry (changes requested):
Looks good to me! The only this is that I just pushed some changes to main which install the Poppins font globally, so you can get rid of the local import here. You can also remove the ``font-family: 'Poppins', sans-serif;`` line since by default now the font is Poppins.

HuxleyBerry (approved):
LGTM

PR #41, opened 2024-12-14
Issue 12 - Created api view and serializer for registering users.

### LINKED ISSUE #12: Create API view for user registration
## Prerequisites

Need the user model to be completed: #3 

## Issue

Create an API view which the frontend can call to register a user. I suggest using function-based views: https://www.django-rest-framework.org/api-guide/views/#function-based-views. The request should provide the following information (these are pretty much the same fields as in the registration modal):
- username
- first name
- last name
- email
- password

The view needs to create an instance of the user model with these fields. The other compulsory fields in the user model can be initialised to sensible defaults. For example, I was thinking that by default, the visibility preference should be "friends only".

someonebear:
## Change Summary
Created api view, and serializer for registering users. Only username, first/last name, email and password are required.

### Change Form

- [X] The change has tests

## Other Information
No :)

# Related issue

- Resolve #12


server/bingo/serializers.py
```diff
@@ -0,0 +1,26 @@
...
+        model = User
+        fields = ['username', 'first_name', 'last_name', 'email', 'password']
+        extra_kwargs = {
+            'password': {'write_only': True},
+            'first_name': {'required': True},
+            'last_name': {'required': True}
+        }
+
+    def create(self, validated_data):
+        user = User(
```
HuxleyBerry: Can you use ``User.objects.create_user()`` instead of ``User()``?

server/bingo/tests.py
```diff
@@ -39,3 +39,24 @@ def test_default_total_completions(self):
...
+
+    def test_register_user(self):
+        response = self.client.post(
+            '/register/', {'username': 'test_user', 'email': 'test@gmail.com',
+                           'first_name': 'larry', 'last_name': 'bird', 'password': 'SuperSecure123'}
+        )
+
+        self.assertEqual(response.status_code, 201)
+
+    def test_register_user_error(self):
```
HuxleyBerry, line 59: Could you add a couple more tests to verify that there's an error if you give an invalid email or an insecure password?

server/api/urls.py
```diff
@@ -17,14 +17,17 @@
...
     TokenRefreshView,
 )
+
 urlpatterns = [
-    path("admin/", admin.site.urls),
+    path('admin/', admin.site.urls),
     path("api/healthcheck/", include(("api.healthcheck.urls"))),
     path('api/token/', TokenObtainPairView.as_view(), name='token_obtain_pair'),
     path('api/token/refresh/', TokenRefreshView.as_view(), name='token_refresh'),
+    path('register/', views.register_user, name='register_user'),
```
HuxleyBerry: Can you change the path to be ``api/register/`` instead? Or even better, move this path to a new ``urls.py`` in the bingo folder and import it to here.

HuxleyBerry (changes requested):
Looks good, just a few things

HuxleyBerry (approved):
LGTM

PR #44, opened 2024-12-14
Added view for user model, urls for bingo and api, unit tests for the…

MahitGtg:
… view

## Change Summary
**views for the user model are set up as per the requirements of the client/src/types/users file**

### Change Form

- [x] The change has tests


server/bingo/tests.py
```diff
@@ -147,3 +150,64 @@ def test_friends_with_self(self):
...
+
+        # Make request to the endpoint
+        response = self.client.get(reverse("current-user"))
+
+        # Check response status code
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        # Check response data matches expected TypeScript interface
+        self.assertEqual(response.data["userId"], self.user.user_id)
+        self.assertEqual(response.data["userName"], "testuser")
```
SafetyInObscurity: The rest of these assertions should match user_id, testing against the value of the model rather than a string

SafetyInObscurity (dismissed):
Looks good, just the tests need a tweak

server/bingo/tests.py
MahitGtg: should be done now!

HuxleyBerry (approved):
LGTM

PR #47, opened 2024-12-14
Issue 4 create navigation bar

### LINKED ISSUE #4: Create navigation bar
Need to wait for https://github.com/codersforcauses/bluecrew/issues/2 to be done. The navbar should be displayed at the top of all pages and should work on both mobile and desktop. It should have links to the following pages:
- The home/bingo page
- The leaderboard page
- The user preferences page
- The friends page

It should also include buttons to log in and to register.

Vuetify has some useful inbuilt components for this. E.g. https://vuetifyjs.com/en/components/toolbars/

mye9:
## Change Summary
Added a NavBar with both desktop and mobile versions, with login and logout status.

### Change Form

- [x] The change has tests

## Other Information
N/A

# Related issue

- Resolve #4


HuxleyBerry (changes requested):
I haven't finished reviewing this yet, but could you please unstage all the files in which no changes have been made? There's a lot of files here that are irrelevant to the issue -- I suspect the line endings got changed which is why you were able to stage them.

client/src/stores/user.ts
```diff
@@ -10,6 +10,11 @@ export const useUserStore = defineStore('user', () => {
 
   const isLoggedIn = computed(() => userData.value !== null)
 
+  // Set Deafult userData to null
+  if (userData.value) {
+    userData.value = null
+  }
```
HuxleyBerry, line 16: I don't think this is necessary since ``userData`` is already set to ``null`` by default in line 7.

client/src/components/NavBarWrapper.vue
HuxleyBerry, line 1: I think you should try and move as much of the logic as possible into this wrapper component. That way we adhere to the DRY principle and do not have the logic repeated in both the mobile and desktop components.

For example, in both versions of the navbar, pressing the sign in button does the same thing. So you should give both versions of the component an emit which the wrapper component listens for. Then you can put the code for responding to the sign in button in one place.

Another example is having the state for whether the user is logged in in the wrapper component and having that state passed down to the mobile and desktop components via a prop.

client/src/components/NavBarMobile.vue
```diff
@@ -0,0 +1,234 @@
...
+  background-color: transparent !important;
+  color: white !important;
+  border: 2px solid white !important;
+  border-radius: 30px !important;
+  height: 48px !important;
+  font-family: 'Lilita One', Arial, sans-serif !important;
+  font-size: 24px !important;
+  font-weight: 400 !important;
+  text-transform: none !important;
+  letter-spacing: 0 !important;
```
HuxleyBerry, line 219: Are all these ``!important``s really necessary? I tried removing them and I didn't notice any affect. This also applies to the other classes which have ``!important``.

client/src/components/NavBarMobile.vue
```diff
@@ -0,0 +1,234 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 0 16px;
+  margin-bottom: 24px;
+}
+
+.guest-mode {
+  color: white;
+  font-family: 'Lilita One', Arial, sans-serif;
```
HuxleyBerry, line 149: The fonts aren't consistent. Let's stick with Poppins. **Edit**: nevermind, I saw that the figma uses both fonts.

client/src/components/NavBarDesktop.vue
```diff
@@ -0,0 +1,146 @@
...
+
+const userStore = useUserStore()
+const modalStore = useModalStore()
+
+const isLoggedIn = computed(() => userStore.isLoggedIn)
+const userName = computed(() => userStore.userData?.userName || '')
+
+const isActive = (name: string) => {
+  return route.name === name
+}
```
HuxleyBerry, line 70: This should be a computed property rather than a function.

client/src/components/NavBarDesktop.vue
```diff
@@ -0,0 +1,146 @@
...
+  justify-content: center;
+  align-items: center;
+  position: fixed;
+  width: auto;
+  height: 104px;
+  top: 52px;
+  right: 55px;
+  max-width: 100%;
+  margin: 0 auto;
+  z-index: 1000;
```
HuxleyBerry, line 105: The positioning and z-index styles here should be removed and instead done in the wrapper component.

client/src/components/NavBarWrapper.vue
HuxleyBerry, line 1: We want the navbar to appear at the top of the page no matter how far the user scrolls down. To achieve that, you can use css along the lines of this: ```.sticky-nav {
  position: fixed;
  left: 0;
  top: 0;
  width: 100%;
  z-index: 999;
}```

client/src/App.vue
```diff
@@ -1,10 +1,16 @@
...
 import { RouterView } from 'vue-router'
 import BluecrewFooter from '@/components/BluecrewFooter.vue'
+import NavBarWrapper from '@/components/NavBarWrapper.vue'
 </script>
 
 <template>
-  <RouterView />
+  <v-app>
+    <NavBarWrapper />
+    <v-main>
```
HuxleyBerry, line 10: We need to make sure that the pages display below the navbar. This depends on what you do elsewhere, but I think one way of achieving this is to add the ``padding-top`` property to the ``v-main`` element (the amount of padding should be the same as the height of the navbar.

client/src/components/NavBarDesktop.vue
```diff
@@ -0,0 +1,146 @@
...
+}
+
+//Navigate Logics to each pages
+const handleNavigation = (page: string) => {
+  if (!isLoggedIn.value) {
+    alert('You must be logged in to access this page.')
+    return
+  }
+  router.push({ name: page })
+}
```
HuxleyBerry, line 79: This component isn't the place to enforce this. We need to implement restricted access to these pages in a more "global" way because the user could also just type the url into the browser's address bar. I will make a separate issue for that. So you can get rid of this function.

client/package.json
```diff
@@ -18,6 +18,7 @@
     "@vueuse/core": "^12.0.0",
     "axios": "^1.7.7",
     "pinia": "^2.2.6",
+    "vant": "^4.9.10",
```
HuxleyBerry, line 21: What is this for?

HuxleyBerry (changes requested):
Great work so far! I have left some comments. The main one is to do some refactoring so that the navbar state and logic is in the wrapper component so that the logic doesn't get duplicated.

client/src/components/NavBarDesktop.vue
```diff
@@ -0,0 +1,146 @@
...
+const truncatedUserName = computed(() => {
+  if (isLoggedIn.value && userName.value.length > 10) {
+    return userName.value.slice(0, 10) + '...'
+  }
+  return userName.value
+})
+</script>
+
+<style scoped>
+@import url('https://fonts.googleapis.com/css2?family=Lilita+One&display=swap');
```
HuxleyBerry, line 91: Could you install this font globally instead like has already been done for Poppins?

HuxleyBerry (changes requested):

Issue #37, opened 2024-12-13
Design landing page


HuxleyBerry:
The client has asked for a simple landing page. They specifically mentioned the duolingo landing page as something they liked. That looks like this (https://www.duolingo.com/):

![image](https://github.com/user-attachments/assets/820faa25-9829-4479-8c25-9dae49ad1f08)

This issue is to design a similar page for the bluecrew bingo game.

survivzhang:
<img width="856" alt="Screenshot 2024-12-14 at 1 22 16 PM" src="https://github.com/user-attachments/assets/c4b00121-e7ca-40dd-81c4-5b8d4fac3855" />

survivzhang:
<img width="881" alt="Screenshot 2024-12-14 at 1 29 06 PM" src="https://github.com/user-attachments/assets/acb41abb-ef02-4562-82c0-cbc48e1c1697" />

HuxleyBerry:
Nice!

PR #53, opened 2025-01-02
Issue 48 make landing page

### LINKED ISSUE #48: Make landing page
The landing page has now been designed in #37, so make the landing page following that wireframe. This page should be written in ``HomeView.vue`` in the ``src/views`` directory. Feel free to rename that file so something like ``LandingPage.vue`` though.

aradavi63:
## Change Summary
Created a landing page view in line with the design specified in issue #37 
The page cannot be fully completed as there is no login modal to assign one of the buttons to, but that is an easy fix that can be done later
Honestly, I'm still not 100% satisfied with the placement of components in the desktop version, but I wanted to put this out here so I can receive feedback on what to change

### Change Form

- [x] The change has tests

## Other Information

# Related issue

- Resolve #48


client/src/router/index.ts
```diff
@@ -9,6 +10,11 @@ const router = createRouter({
       name: 'home',
       component: HomeView,
     },
+    {
+      path: '/landing',
+      name: 'landing',
+      component: LandingView,
+    },
```
HuxleyBerry: Can you change the path to be ``'/'`` rather than ``'/landing'``?

client/src/views/LandingView.vue
```diff
@@ -0,0 +1,171 @@
...
+              rounded="xl"
+              size="x-large"
+              @click="openRegisterButton"
+            >
+              I already have an account
+            </v-btn>
+          </div>
+        </div>
+      </div>
+    </template>
```
HuxleyBerry: I think you should be able to avoid having two entirely separate templates for the mobile and desktop versions of the page. Their appearances are very similar, with the only big difference being that in the mobile version all the elements are stacked in a single column. I believe you will be able to achieve this effect in a more elegant manner using Vuetify's grid system (which includes breakpoints for mobile): https://vuetifyjs.com/en/components/grids/#grid-system.

The benefit of this is you won't be repeating yourself so much.

client/src/views/LandingView.vue
```diff
@@ -0,0 +1,171 @@
...
+            </v-btn>
+            <v-btn
+              block
+              class="bg-primaryBlue text-creamyWhite"
+              rounded="xl"
+              size="x-large"
+              @click="openRegisterButton"
+            >
+              I already have an account
+            </v-btn>
```
HuxleyBerry: The text doesn't fit on the button for me on mobile.

client/src/views/LandingView.vue
```diff
@@ -0,0 +1,171 @@
...
+          </div>
+        </div>
+      </div>
+    </template>
+
+    <RegisterModal v-if="modalStore.currentModal === 'register'" @close="modalStore.closeModal" />
```
HuxleyBerry: You should move this usage of the RegisterModal component into ``App.vue``. That way, opening the registration modal will work from any page.

HuxleyBerry (changes requested):
Thanks for working on this over the break! It's looking almost done -- I just had a few comments.

client/src/views/LandingView.vue
```diff
@@ -0,0 +1,171 @@
...
+      </div>
+    </template>
+
+    <!-- Mobile Version -->
+    <template v-else>
+      <div class="mobile-overlay">
+        <div class="mobile-content">
+          <div class="mobile-close">
+            <v-icon>mdi-close</v-icon>
+          </div>
```
HuxleyBerry: Not your fault since the figma had this close icon, but I don't think there's any reason for it (?), so I think it should be removed. Similarly, I don't think there's any need for the ``.mobile-overlay`` div.

client/src/views/LandingView.vue
```diff
@@ -0,0 +1,171 @@
...
+            </v-btn>
+            <v-btn
+              id="login-button"
+              class="bg-primaryBlue text-creamyWhite"
+              rounded="xl"
+              size="x-large"
+              @click="openRegisterButton"
+            >
+              I already have an account
+            </v-btn>
```
HuxleyBerry: The text and the buttons should be vertically centred and have more spacing between each other as per the figma.

HuxleyBerry (changes requested):

client/src/views/LandingView.vue
aradavi63: It looks like this for me, but I will change it so that there is more margin on the sides 
Screenshot 2025-01-03 at 10 32 11 pm

client/src/App.vue
```diff
@@ -1,11 +1,17 @@
...
+import RegisterModal from '@/components/RegisterModal.vue'
+
+const modalStore = useModalStore()
 </script>
 
 <template>
   <RouterView />
   <BluecrewFooter />
+  <RegisterModal />
+  <RegisterModal v-if="modalStore.currentModal === 'register'" @close="modalStore.closeModal" />
```
HuxleyBerry: The RegisterModal shouldn't be here twice.

client/src/router/index.ts
```diff
@@ -1,14 +1,20 @@
...
 
 const router = createRouter({
   history: createWebHistory(import.meta.env.BASE_URL),
   routes: [
     {
-      path: '/',
+      path: '/home',
       name: 'home',
       component: HomeView,
     },
```
HuxleyBerry: Now that you've got the landing page, you should delete this route.

client/src/views/LandingView.vue
```diff
@@ -0,0 +1,81 @@
...
+          <v-btn
+            id="login-button"
+            class="bg-primaryBlue text-creamyWhite"
+            rounded="xl"
+            size="x-large"
+            @click="openRegisterButton"
+          >
+            I already have an account
+          </v-btn>
+        </div>
```
HuxleyBerry: I think these buttons should have some margin on the left and right

HuxleyBerry (changes requested):

HuxleyBerry (approved):
Looks good, we can finish the small centring problem in a later issue.

PR #60, opened 2025-01-04
add bingo grid model and tests

### LINKED ISSUE #57: Create db model for the bingo grid
The model should have the following fields:
- grid_id (primary key)
- 16 challenge fields. However, instead of creating 16 individual fields on this model, I think we can use an ``ArrayField`` since we are using postgres: https://docs.djangoproject.com/en/5.1/ref/contrib/postgres/fields/#arrayfield. I've never used it myself, but it looks good from a skim of the docs.
- is_active (boolean). Also implement some validation which ensures that at most one bingo grid is active at one time.

notalex-sh:
## Change Summary
Added bingo grid model and unit tests

### Change Form

- [x] The change has tests

## Other Information

# Related issue

- Resolve #57


server/bingo/models.py
```diff
@@ -137,3 +138,28 @@ def clean(self):
...
         return f"Friend request from {self.requester} to {self.receiver} ({self.status.capitalize()})"
+
+
+class BingoGrid(models.Model):
+    #  A Bingo grid that holds an array of 16 challenge IDs.
+    grid_id = models.AutoField(primary_key=True)
+    # We store an array of integers (challenge IDs).
+    challenges = ArrayField(
+        base_field=models.IntegerField(),
+        size=16,
```
HuxleyBerry: Can you add some validation so that there have to be exactly 16 challenges?

server/bingo/models.py
```diff
@@ -137,3 +138,28 @@ def clean(self):
...
     def __str__(self):
         return f"Friend request from {self.requester} to {self.receiver} ({self.status.capitalize()})"
+
+
+class BingoGrid(models.Model):
+    #  A Bingo grid that holds an array of 16 challenge IDs.
+    grid_id = models.AutoField(primary_key=True)
+    # We store an array of integers (challenge IDs).
+    challenges = ArrayField(
+        base_field=models.IntegerField(),
```
HuxleyBerry: On reflection, I'm a little worried about using an ``ArrayField`` here. We want to ensure that all of the ids stored are ids of currently existing challenges, but just using an integer field doesn't guarantee that. Maybe a better way is to use a ``ManyToMany`` field, though the problem with that is we lose the order of the challenges. I think there are some solutions to that problem here: https://stackoverflow.com/questions/1880530/django-many-to-many-insert-ordering

HuxleyBerry (changes requested):

HuxleyBerry (approved):
LGTM

PR #69, opened 2025-01-04
Issue 21 make api view for retrieving leaderboard

### LINKED ISSUE #21: Make API view for retrieving leaderboard
## Prerequisites
#3 

## Issue
Make an API view which retrieves the top twenty (maybe subject to change) users according to their total points. It just needs to retrieve the name, rank, and points of each of those ten users (in order). You can use Django's ``order_by()`` method for this: https://docs.djangoproject.com/en/5.1/ref/models/querysets/#order-by.

Users with the same number of points should be given the same rank, so you'll have to do a bit more work than just using ``order_by``.

Additionally, the view should return the rank of the currently logged in user, since we'll also want to display that on the leaderboard page even if they aren't in the top twenty.

bibonut:
## Change Summary

### Change Form

- [X] The change has tests

## Other Information


server/api/settings.py
```diff
@@ -58,6 +57,10 @@
 # Change this to point to custom user model
 AUTH_USER_MODEL = "bingo.User"
 
+SIMPLE_JWT = {
+    'USER_ID_FIELD': 'user_id',
+    "ACCESS_TOKEN_LIFETIME": timedelta(minutes=60),
+}
```
HuxleyBerry: There's already a ``SIMPLE_JWT`` configuration in ``settings.py``, so this is unnecessary.

server/bingo/tests/test_leaderboard_view.py
```diff
@@ -0,0 +1,53 @@
...
+        self.user2 = User.objects.create_user(
+            username="user2", email="user2@example.com", password="password", total_points=1)
+        self.user3 = User.objects.create_user(
+            username="user3", email="user3@example.com", password="password", total_points=3)
+
+        self.client = Client()
+
+    def test_leaderboard(self):
+        # GET leaderboard api
+        leaderboard_response = self.client.get("/api/leaderboard", follow=True)
```
HuxleyBerry: You can get rid of the ``follow=True`` if you add a slash to the end of the path. Alternatively, you can use ``django.urls.reverse`` and do something like ``leaderboard_response = self.client.get(reverse('get_leaderboard'))``.

server/bingo/tests/test_leaderboard_view.py
```diff
@@ -0,0 +1,53 @@
...
+        leaderboard_response_content = json.loads(leaderboard_response.content.decode("utf8"))
```
HuxleyBerry: You can do this line more simply: ``leaderboard_response_content = leaderboard_response.data``.

server/bingo/tests/test_leaderboard_view.py
```diff
@@ -0,0 +1,53 @@
...
+
+        self.assertEqual(leaderboard_response_content[0]["username"], "user3")
+        self.assertEqual(leaderboard_response_content[0]["total_points"], 3)
```
HuxleyBerry: Can you change these lines to use the values defined in the ``setUp`` function? So for example, the first line should be ``self.assertEqual(leaderboard_response_content[0]["username"], self.user3.username)``. The same goes for the similar lines below.

server/bingo/tests/test_leaderboard_view.py
```diff
@@ -0,0 +1,53 @@
...
+class LeaderboardTest(TestCase):
+    def setUp(self):
+        self.user1 = User.objects.create_user(
+            username="user1", email="user1@example.com", password="password", total_points=1)
+        self.user2 = User.objects.create_user(
+            username="user2", email="user2@example.com", password="password", total_points=1)
+        self.user3 = User.objects.create_user(
+            username="user3", email="user3@example.com", password="password", total_points=3)
+
+        self.client = Client()
```
HuxleyBerry: Use ``rest_framework.test.APIClient`` instead of ``django.test.Client``.

server/bingo/tests/test_leaderboard_view.py
```diff
@@ -0,0 +1,53 @@
...
+            "/api/token/", {"username": "user1", "password": "password"}, content_type="application/json")
+        token_response_content = json.loads(
+            token_response.content.decode('utf-8'))
+
+        # GET leaderboard api
+        user_token = token_response_content["access"]
+        leaderboard_response = self.client.get(
+            "/api/leaderboard", headers={"Authorization": "Bearer " + user_token}, follow=True)
+        leaderboard_response_content = json.loads(
+            leaderboard_response.content.decode("utf8"))
```
HuxleyBerry: You can avoid having to do all this stuff with manually adding an authorisation token to the headers by doing ``self.client.force_authenticate(user=self.user1)`` (where ``self.client`` is a ``rest_framework.test.APIClient``).

HuxleyBerry (changes requested):

HuxleyBerry (approved):
LGTM

PR #91, opened 2025-01-10
Issue 62 create api view for fetching bingo grid

### LINKED ISSUE #62: Create API view for fetching bingo grid
## Prerequisites
- #57 and #58 

## Issue
- This view needs to fetch all of the challenge information for the currently active bingo grid (probably in the form of an array of length 16)
- Additionally, it needs to fetch the completion status of all these challenges from the ``GridInteraction`` model
- There may not exist an instance of the ``GridInteraction`` model with the corresponding user and grid id. If so, then an instance should be created
- However, if the user is not logged in, then there is no need to fetch completion statuses or create an instance of the ``GridInteraction`` model.

someonebear:
## Change Summary
Created view to get active bingo grid and the challenges that it contains. Added tests for this view.
The view will also return the completion status of the challenges in the bingo grid for the user, if the user is authenticated.
If a TileInteraction object exists for the challenge and grid, and if the completion status for that TileInteraction is true, then the challenge is displayed as completed, otherwise as started. If no TileInteraction object exists, the challenge is displayed as not started.

### Change Form

- [X] The change has tests

## Other Information

# Related issue

- Resolve #62


HuxleyBerry (approved):
Nice work, LGTM!

PR #96, opened 2025-01-11
Issue 50 achieve style consistency between login and registration modals

### LINKED ISSUE #50: Achieve style consistency between login and registration modals
Since they were made by different teams, the registration and login modals don't look quite the same stylistically. For example, at the time of writing, the buttons are different heights and the close buttons are different. This issue is about resolving the style differences. Ideally make them both look like the figma.

samueltl21:
## Change Summary

### Change Form

- [ ] The change has tests

## Other Information

# Related issue

- Resolve #50

### DIFF
--- a/client/src/components/LoginModal.vue
+++ b/client/src/components/LoginModal.vue
@@ -51,7 +51,7 @@ const valid = ref(false)
       </v-card-title>
 
       <v-card-subtitle class="text-center subtitle mt-2 text-primaryPink">
-        <h3><b>Welcome Back</b></h3>
+      <strong><h3><b>Welcome Back</b></h3></strong>
       </v-card-subtitle>
       <v-card-text>
         <p class="text-center subtitle mb-4 text-primaryPink">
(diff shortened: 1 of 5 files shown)


client/src/components/LoginModal.vue
```diff
@@ -89,7 +97,7 @@ const valid = ref(false)
 
           <div class="mt-3">
             <a href="/forgot-password" class="text-lightBlue">
-              <b class="text-subtitle-2">Forgot Password?</b>
+              <b class="text-subtitle-2"><forgotpass>Forgot Password?</forgotpass></b>
```
HuxleyBerry: I get an error here because ``forgotpass`` isn't a valid element. Use a ``<p></p>`` instead.

HuxleyBerry (changes requested):
Nice work. I just noticed two small problems. Also, I think these modals would look better if they were 500px wide rather than 400px.

client/src/components/RegisterModal.vue
```diff
@@ -49,7 +53,9 @@ const submitForm = () => {
       <v-card>
         <v-card-text style="height: auto; overflow-y: auto">
           <div class="header">
-            <button class="close-button" @click="closeDialog">x</button>
+            <button class="close-button" @click="closeDialog">
+              <v-icon icon="mdi-close-circle-outline" class="mr-3 mt-3"></v-icon>
+            </button>
```
HuxleyBerry, line 58: These icons are not in the same place in both modals.

HuxleyBerry (approved):
LGTM

PR #98, opened 2025-01-11
Issue 56 create api view for marking a challenge as started

### LINKED ISSUE #56: Create API View for marking a challenge as started
## Prerequisites
#52 

## Issue
Make an API view which the frontend can call to mark a challenge as started by the user. It should take a number from 0-15 and create a ``TileInteraction`` instance with that number as the ``position``, the id of the currently logged in user, and the id of the currently active bingo grid.

Also write some unit tests.

SafetyInObscurity:
## Change Summary
Created the view to start a challenge on the grid

### Change Form
- [x] The change has tests

# Related issue

- Resolve #56


server/bingo/urls.py
```diff
@@ -14,6 +14,7 @@
     path('user/me/', views.get_current_user, name='current_user'),
     path('delete-friendship/<int:friendship_id>/',
          views.delete_friendship, name='delete_friendship'),
+    path('challenge', views.start_challenge, name="start-challenge"),
```
HuxleyBerry: Can you change the name to have an underscore instead of a hyphen to be consistent. Also, can you change the path to have "start" in the name and have a trailing slash?

HuxleyBerry (changes requested):

HuxleyBerry (approved):

Issue #76, opened 2025-01-06
Make bingo page


HuxleyBerry:
## Prerequisites
#54 and #55 

## Issue
Use the components from issues #54 and #55 to make the bingo page. Here's what the page should look like on desktop:
![image](https://github.com/user-attachments/assets/40795d18-1a7d-4780-a799-b61bd6e3ce2c)

...and on mobile:

![image](https://github.com/user-attachments/assets/844fdb38-fd55-4aee-9883-68fbddd65620)

Note that the challenge component (from #55) will have to be placed in a dialog in the mobile version

Also, make sure the contents of the challenge component is updated depending on which of the bingo tiles is selected.

aradavi63:
So the challenge component will not be in a dialog for the desktop version? Is it opening a new view?

HuxleyBerry:
> So the challenge component will not be in a dialog for the desktop version? Is it opening a new view?

Correct, it won't be in a dialog. What I mean by that is that on desktop, you will be able to interact with the rest of the screen (such as tapping a new bingo tile) without needing to first close the challenge component. On the other hand, on mobile, you won't be able to interact with the website until you dismiss the dialog that the challenge component is in,

PR #106, opened 2025-01-13
Issue 54 create component for bingo tile

### LINKED ISSUE #54: Create component for bingo tile
Make a "bingo tile" vue component. They should look something like this (screenshot from the figma):

image

The component should have the following props:
- The type (connect/understand/act) which will determine the icon shown
- The text to be shown on the tile. I suspect there isn't going to be enough room on mobile to fit the text if it's reasonably sized, so maybe the text shouldn't be shown on mobile. I suggest experimenting with this.
- The status (not started/started/completed), which determines the colour of the tile
- A boolean prop to reflect whether the tile is currently selected. If true, the tile should be highlighted/emphasised in some way.

melbauerr:
## Change Summary
Create bingo tile component with options for the different props:
- type (connect/understand/act) determines icon shown
- text is same on mobile and desktop, on mobile text is only sufficient for 3 words
- status (not started/started/completed), determines colour of the tile
- boolean prop reflects whether tile is currently selected - if true, tile is emphasised
Screenshot 2025-01-13 at 12 46 01 pm

### Change Form
- [❌] The pull request title has an issue number
- [❌] The change works by "Smoke testing" or quick testing
- [ ] The change has tests

## Other Information
Pull request also includes placeholder Blingo tile page.

# Related issue

- Resolve #54


client/src/components/BingoTile.vue
```diff
@@ -0,0 +1,97 @@
...
+    case 'bg-creamWhite':
+      return 'dark'
+    default:
+      return 'light'
+  }
+})
+</script>
+
+<template>
+  <div
```
HuxleyBerry, line 53: Can you add the ``cursor-pointer`` class to this div? See https://vuetifyjs.com/en/styles/cursor/#usage.

client/public/one-man.svg
HuxleyBerry, line 1: You don't seem to be using this file, so you can delete it

client/public/man.svg
HuxleyBerry, line 1: Same with this file

HuxleyBerry (changes requested):
Looks good, I've just pointed out some minor things. Additionally, I noticed the tiles have a fixed size which makes them too small on desktop. Instead, they should take up the full width and height of their parent. However, that can be addressed in issue #76 instead if you like.

HuxleyBerry (approved):
LGTM

PR #107, opened 2025-01-13
Issue 78 make friends components

### LINKED ISSUE #78: Make friends components
Make the components that will be displayed on the friends pages. There are a few different variants:
- A simple variant with nothing on the right
image
- A variant which allows you to accept or reject
image
- A variant with a "dimiss" button:
image
- A variant with a "delete" button. This should look the same as the dismiss button variant, except that it says "delete" rather than "dismiss". Also, a popup (use a v-dialog with a v-card) should appear asking the user to confirm that they want to delete.

I think a good way of going about having these different variants would be to first make a "base component" which has a [slot](https://vuejs.org/guide/components/slots) for the elements on the right, and then to make a component for each of the variants by using the base component with the appropriate thing in the slot.

aradavi63:
## Change Summary
Made the friends component, not sure if I used the slots properly so feel free to suggest improvement there

### Change Form
May need to work on the dialog box for deletion of a friend

- [ ] The change has tests

# Related issue

- Resolve #78


client/src/router/index.ts
```diff
@@ -34,7 +35,7 @@ const router = createRouter({
     {
       path: '/blingo',
       name: 'blingo',
-      component: PlaceholderView,
+      component: FriendView,
```
HuxleyBerry: Can you make the ``/friends`` path point to this view instead?

client/src/components/FriendComponent.vue
```diff
@@ -0,0 +1,109 @@
...
+  variant?: 'default' | 'acceptReject' | 'dismiss' | 'delete' | 'addFriend' | 'accept'
+}>()
+
+const showDialog = ref(false)
+
+const emit = defineEmits(['accept', 'reject', 'dismiss', 'delete', 'addFriend'])
+</script>
+
+<template>
+  <v-col class="rounded-lg d-flex ga-3 px-5 outline align-center bg-creamWhite text-primaryGrey">
```
HuxleyBerry: I don't think the outermost element of a component should be a ``v-col`` because a ``v-col`` is a "layout component" which supposed to be used within other layout components (v-row, v-container). Components are supposed to be agnostic about where they're used, so using a ``v-col`` here goes against that somewhat. Can you try changing it to a div?

I noticed the ``LeaderboardRow`` component also makes this "mistake" -- maybe you can fix that one too?

client/src/components/FriendComponent.vue
```diff
@@ -0,0 +1,109 @@
...
+      </slot>
+    </div>
+
+    <!-- Dismiss variant -->
+    <div v-if="variant === 'dismiss'" class="dismiss-action">
+      <slot name="dismiss">
+        <v-btn color="primaryBlue" variant="flat" @click="emit('dismiss')" class="dismiss-btn">
+          Dismiss
+        </v-btn>
+      </slot>
```
HuxleyBerry: As you alluded to, you aren't using slots correctly. If you delete all the slot tags in this component, the component functions exactly the same. Once you've done that, this component won't be using slots at all, but what you've done is also a perfectly valid method. Using slots was just a suggestion of mine.

HuxleyBerry (changes requested):
Nice work! I've requested a few changes.

aradavi63:
Hi Huxley, I changed the v-col to div as you requested, but I am unsure of how to get the styling right, as the custom colour themes we have set in main.ts don't seem to be applying to the divs
image

In addition, how do I test the view while having Auth, what I did was temporarily remove the "meta: { requiresAuth: true }," line until this issue is done, after which I will return it. Thanks.

HuxleyBerry:
> Hi Huxley, I changed the v-col to div as you requested, but I am unsure of how to get the styling right, as the custom colour themes we have set in main.ts don't seem to be applying to the divs image

To get the colours to work in the style, you'll need to do something like ``background-color: rgb(var(--v-theme-primaryBlue));``. I think you'll also need to add some padding to get the component to look like what you had before.

HuxleyBerry:
>In addition, how do I test the view while having Auth, what I did was temporarily remove the "meta: { requiresAuth: true }," line until this issue is done, after which I will return it. Thanks.

Instead, can you only make that change to ``index.ts`` locally? So revert your changes to ``index.ts``, commit that, and then you can make the change again, but don't commit that change.

HuxleyBerry (approved):

PR #109, opened 2025-01-14
added the error page and encoporated it in the router

### LINKED ISSUE #102: Create custom 404 page
Create a custom 404 page which says something along the lines of "page not found" as has a "return home" button which takes the user back to the landing page. There is no figma design for this page currently, so making one could be a good first step.

MahitGtg:
## Change Summary
- Created an error page called 404Error.view showing people an error line with 'Go Home' button
- Added router settings so that people are redirected to this page if they are trying to visit a route not already made.
### Change Form

- [ x ] The pull request title has an issue number
- [ ] The change has tests
- [ x ] The change has documentation

## Other Information
Screenshot 2025-01-15 at 1 04 0
3 AM
Screenshot 2025-01-15 at 1 04 20 AM

# Related issue

- Resolve #102


client/src/views/404Error.vue
```diff
@@ -0,0 +1,69 @@
...
+          </div>
+        </div>
+      </v-col>
+    </v-row>
+  </v-container>
+</template>
+
+<style scoped>
+.error-container {
+  height: calc(100vh - 144px);
```
HuxleyBerry: If you replace this by ``height: 100%``, you get rid of the vertical scroll bar on desktop. At least, that's what I observed testing on my computer in Chrome.

HuxleyBerry (approved):
Nice, I like it! If you want to add a bit more to it, you could add "404" in big letters at the top and maybe an icon (I'm thinking a submarine could be a cool option). But it already looks good as it is.

client/src/views/404Error.vue
MahitGtg: Ive tried a little something, let me know how it looks.
Not sure about it honestly cause im not that good with design

PR #125, opened 2025-01-18
Issue 95 create api view to request friendship

yuwengchoong:
## Change Summary

### Change Form

- [ YES] The pull request title has an issue number
- [YES ] The change has tests

## Other Information


server/bingo/tests/test_friendship_request.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def test_request_friendship_to_self(self):
+        self.client.force_authenticate(user=self.user1)
+        request_friendship_url = reverse('request_friendship', args=[self.user1.user_id])
+        response = self.client.post(request_friendship_url)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.data["error"], "You cannot send a friendship request to yourself.")
+        with self.assertRaises(ValidationError):
+            Friendship.objects.create(
+                requester=self.user1, receiver=self.user1).full_clean()
```
HuxleyBerry: This isn't necessary since it is already tested in the friendship model tests

HuxleyBerry (approved):
Nice work! I've just pointed out one unnecessary thing you included in your tests, but otherwise everything looks good.

server/bingo/views.py
```diff
@@ -180,6 +180,34 @@ def get_leaderboard(request):
...
+        )
+        return Response(
+            {"message": "Friendship request sent successfully."},
+            status=status.HTTP_201_CREATED
+        )
+    except IntegrityError:
+        return Response(
+            {"error": "A friendship request already exists or is pending."},
+            status=status.HTTP_409_CONFLICT
+        )
```
HuxleyBerry, line 205: Sorry to point out another thing which I should've noticed in person earlier, but this doesn't cover the case where a "reverse friendship request" exists. That is, the situation where user X is trying to use your view to request the friendship of user Y, but there already exists an instance of the ``Friendship`` model in the db which has user X as the *receiver* and Y as the *requester*. In this case, the view should respond with an error.

 This logic is actually already implemented in the Friendship model's ``clean()`` function in ``models.py``. It would be good to avoid duplicating that code. Unfortunately, django doesn't automatically run the clean function when you call the ``objects.create()`` method. To run that clean function, you'll need to replace ``Friendship.objects.create(...)`` with something like 
```
new_friendship = Friendship(requester=request.user, receiver=receiver, status=Friendship.PENDING) # this doesn't save to the db
new_friendship.full_clean() # this is the line which will run the validation code in the model's clean() function
new_friendship.save()
```
Once you make that change, you can also remove the code you've written for checking for self-requests since that is also already handled in the Friendship model's clean function.

HuxleyBerry (changes requested):

server/bingo/views.py
```diff
@@ -180,6 +180,37 @@ def get_leaderboard(request):
...
+@api_view(['POST'])
+@permission_classes((permissions.IsAuthenticated, ))
+def request_friendship(request, user_id):
+    receiver = get_object_or_404(User, user_id=user_id)
+
+    if request.user == receiver:
+        return Response(
+            {"error": "You cannot send a friendship request to yourself."},
+            status=status.HTTP_400_BAD_REQUEST
+        )
```
HuxleyBerry: You can get rid of this since this logic is handled by the clean() function in the Friendship model.

server/bingo/views.py
```diff
@@ -180,6 +180,37 @@ def get_leaderboard(request):
...
+        new_friendship.full_clean()
+        new_friendship.save()
+        return Response(
+            {"message": "Friendship request sent successfully."},
+            status=status.HTTP_201_CREATED
+        )
+    except ValidationError as e:
+        return Response(
+            {"error": e.message_dict.get('__all__', ['Validation error'])[0]},
+            status=status.HTTP_409_CONFLICT
```
HuxleyBerry: I think a 400 error would be more suitable here.

HuxleyBerry (changes requested):

HuxleyBerry (approved):
Nice, LGTM

PR #127, opened 2025-01-22
Issue 75 create api view for completing a challenge

### LINKED ISSUE #75: Create API view for completing a challenge
## Prerequisites
#57 and #58

## Issue
This issue is for creating the view which is called when the "Finish" button is pressed:

image

The view should take the following arguments:
- A number in the range 0-15 representing the index of the bingo tile which is being completed
- A boolean representing whether the user has consented to bluecrew uploading the picture to social media
- An image

The view should then do the following:

1. Get the instance of the ``TileInteraction`` model corresponding to the combination of the logged in user, the current bingo grid, and the supplied index
2. Update the ``consent`` and ``image`` fields of the ``TileInteraction`` instance and mark it as completed.
3. Award the user the amount of points the challenge was worth
4. Now comes the complicated part...calculate if any bingos (row, column or diagonal) are achieved by the completion of this challenge.
5. For each bingo, award the user 100 points (not sure about 100, maybe make it a constant in ``settings.py``)
6. If the whole grid is completed, award the user 500 points (again make it a constant in ``settings.py``)
7. Return a message to the frontend containing:
- The amount of points the challenge was worth
- A list of all the rows/columns/diagonals in which a bingo was achieved (you'll have to come up with some representation)
- Whether the whole grid just got completed

someonebear:
## Change Summary
Added a view to complete a challenge, and give the user any points they've gotten from completing the challenge, as well as any bingos they have achieved.

### Change Form

- [X] The change has tests

## Other Information
: )

# Related issue

- Resolve #75


server/bingo/views.py
```diff
@@ -259,3 +266,53 @@ def get_bingo_grid(request):
...
+    tile = get_object_or_404(TileInteraction, user=request.user,
+                             grid=active_grid, position=serializer.data['position'])
+
+    # Update consent field.
+    tile.consent = serializer.data['consent']
+
+    # Image is not a required field.
+    image = serializer.data.get('image', None)
+    if image and tile.consent:
+        tile.image = image
```
HuxleyBerry: You don't need to check for consent here. The point of the consent field is to track whether the user gives permission for the image to be shared on Blue Crew's social media. So the image the user uploads should be stored regardless.

server/bingo/views.py
```diff
@@ -259,3 +266,53 @@ def get_bingo_grid(request):
...
+    'challenge_points' contains the points earned from completing the challenge.
+    'bingo_points' contains the points earned from bingos completed
+    'bingo_rows' contains the row in which a bingo was just achieved, if any, from 0-3, -1 if no bingo
+    'bingo_cols' contains the column in which a bingo was just achieved, if any, from 0-3, -1 if no bingo
+    'bingo_diag' contains the diagonal in which a bingo was just achieved, if any, denoted by the first row
+    tile the diagonal contains, either 0 or 3, -1 if no bingo.
+    'full_bingo' contains a boolean, representing whether the full grid has been completed.
+    """
+    active_grid = get_object_or_404(BingoGrid, is_active=True)
+    serializer = ChallengeCompleteSerializer(request.data)
```
HuxleyBerry: I think you'll want to call ``serializer.is_valid()``. Otherwise, there isn't much point of a serializer -- you could just use ``request.data`` everywhere you are using ``serializer.data``.

server/bingo/views.py
```diff
@@ -259,3 +266,53 @@ def get_bingo_grid(request):
...
+    This view returns a dictionary the following information.
+    'challenge_points' contains the points earned from completing the challenge.
+    'bingo_points' contains the points earned from bingos completed
+    'bingo_rows' contains the row in which a bingo was just achieved, if any, from 0-3, -1 if no bingo
+    'bingo_cols' contains the column in which a bingo was just achieved, if any, from 0-3, -1 if no bingo
+    'bingo_diag' contains the diagonal in which a bingo was just achieved, if any, denoted by the first row
+    tile the diagonal contains, either 0 or 3, -1 if no bingo.
+    'full_bingo' contains a boolean, representing whether the full grid has been completed.
+    """
+    active_grid = get_object_or_404(BingoGrid, is_active=True)
```
HuxleyBerry: Not a big deal, but I think it would make more sense to return a 500 error here. If there no active BingoGrid, then something has gone very wrong on the server's side.

server/bingo/views.py
```diff
@@ -259,3 +266,53 @@ def get_bingo_grid(request):
...
+    tile.completed = True
+    tile.date_completed = timezone.now()
+    tile.save()
+
+    challenge = active_grid.challenges.all()[tile.position]
+    challenge.total_completions += 1
+    challenge.save()
+
+    user = get_object_or_404(User, is_active=True,
+                             user_id=tile.user.user_id)
```
HuxleyBerry: Does ``request.user`` work instead here?

HuxleyBerry (changes requested):
Great work! I've requested a few small changes, but other than that it all looks perfect to me.

HuxleyBerry (approved):
LGTM!

PR #135, opened 2025-01-25
Issue 116 create user preferences page

### LINKED ISSUE #116: Create user preferences page
Create the user preferences page. Initially, the page should look something like this:

Image

Then once the user has pressed the edit button, the page should change its appearance to something like this:

**Edit**: oops wrong image

See the figma for what the page should look like on desktop. I suggest replacing the text which says "Home >> User Preferences >> Edit Profile" with simply "Preferences" and making that appear in both states (i.e. editing and viewing) of the page.

Also, if the user edits their avatar or bio, the values in the user store should be modified appropriately. That should cause the user's displayed avatar and bio to change too. In the future, we'll need to check with the backend first before making those changes, but that is outside the scope of this issue, so for now just leave a comment in the spot in the code that would need to be done.

aradavi63:
## Change Summary
Added a preference page, which can switch between the viewing mode or the editing mode
- Editing mode allows the user to change their avatar, bio and account visibility
- Has a desktop and mobile view
- Added the OceanYouth teal green as `primaryGreen`
-
### Change Form

- [ ] The change has tests

## Other Information
- Need to connect information to the user store

# Related issue

- Resolve #116


client/src/views/PerferenceView.vue
HuxleyBerry, line 1: You misspelt the name of this file :)

client/src/views/PerferenceView.vue
```diff
@@ -0,0 +1,226 @@
...
+  {
+    src: 'https://t4.ftcdn.net/jpg/00/64/67/63/360_F_64676383_LdbmhiNM6Ypzb3FM4PPuFP9rHe7ri8Ju.jpg',
+  },
+  {
+    src: 'https://t4.ftcdn.net/jpg/00/64/67/63/360_F_64676383_LdbmhiNM6Ypzb3FM4PPuFP9rHe7ri8Ju.jpg',
+  },
+  {
+    src: 'https://t4.ftcdn.net/jpg/00/64/67/63/360_F_64676383_LdbmhiNM6Ypzb3FM4PPuFP9rHe7ri8Ju.jpg',
+  },
+]
```
HuxleyBerry: Some actual avatar files are in main now, so feel free to use those instead.

client/src/views/PerferenceView.vue
```diff
@@ -0,0 +1,226 @@
...
+
+            <!-- Bio Section -->
+            <v-row>
+              <v-col cols="12">
+                <p class="text-h6 font-weight-bold mb-4">Bio</p>
+                <v-textarea
+                  v-model="bio"
+                  variant="outlined"
+                  bg-color="rgb(var(--v-theme-primaryBrown))"
+                ></v-textarea>
```
HuxleyBerry: Add a placeholder here which says something like "Add your bio here". I was sorta wrong about what I said about this textarea in person. You can keep the v-model binding -- it's just if the user doesn't already have a bio that the placeholder will appear.

client/src/views/PerferenceView.vue
```diff
@@ -0,0 +1,226 @@
...
+                  selected-class="bg-primaryGreen"
+                  v-model="visibility"
+                >
+                  <v-btn class="mb-2 bg-primaryBrown" variant="outlined" value="Bluecrew only"
+                    >Bluecrew only</v-btn
+                  >
+                  <v-btn class="mb-2 bg-primaryBrown" variant="outlined" value="Friend only"
+                    >Friend only</v-btn
+                  >
+                  <v-btn class="bg-primaryBrown" variant="outlined" value="Public">Public</v-btn>
```
HuxleyBerry: You need to show which of these is currently selected in some way.

client/src/views/PerferenceView.vue
```diff
@@ -0,0 +1,226 @@
...
+                    max-width="96"
+                    min-width="96"
+                    cover
+                    class="rounded-circle cursor-pointer"
+                    :class="{ 'border-primary': selectedAvatar === index }"
+                    @click="selectedAvatar = index"
+                  />
+                </div>
+              </v-col>
+            </v-row>
```
HuxleyBerry: These avatars should be centred horizontally.

HuxleyBerry (changes requested):
Reminding you of the changes we discussed today :)

client/src/views/PerferenceView.vue
```diff
@@ -0,0 +1,226 @@
...
+        </v-col>
+        <v-col cols="12">
+          <v-btn
+            class="bg-primaryBlue"
+            prepend-icon="mdi-pencil"
+            :block="xs"
+            @click="handleEditClick"
+          >
+            Edit Profile
+          </v-btn>
```
HuxleyBerry: There should be more space below this button (a margin probably would do it).

client/src/views/PerferenceView.vue
aradavi63: Are these the svg files, such as jellyfish.svg and clown-fish.svg?

HuxleyBerry: Yep, see ``utils/avatar.ts``

aradavi63: image
I managed to get these 7 avatars, but when I try to add any more such as `submarine.svg` I get a weird glitch where some of the options go invisible
image
Here, the submarine and whale disappear

HuxleyBerry: You'll have to push your changes if you want me to have a look. In any case, there should be only 6 avatars.

client/src/views/PreferenceView.vue
```diff
@@ -0,0 +1,227 @@
...
+  {
+    src: 'stingray.svg',
+  },
+  {
+    src: 'turtle.svg',
+  },
+  {
+    src: 'jellyfish.svg',
+  },
+]
```
HuxleyBerry, line 38: This list of avatar paths shouldn't be included here since we already have the list of paths in ``utils/avatar.ts``, and it would be bad to have that information duplicated since that could lead to inconsistency. Instead they should be imported from ``utils/avatar.ts``.

HuxleyBerry (approved):
Nice work, LGTM! I pointed out one thing, but that can be fixed in the connecting to backend issue.

PR #142, opened 2025-01-25
Issue 140 use router on login and logout

### LINKED ISSUE #140: Use router on login and logout
The ``logout()`` function in the user store should navigate the user back to the home page if they are currently on a protected route

The ``login()`` function in the user store should "reload" the current page.

HuxleyBerry:
## Change Summary
- I've made it so that when the user logs in or out, the router-view component is rerendered (I followed the advice [here](https://medium.com/hackernoon/the-correct-way-to-force-vue-to-re-render-a-component-bde2caae34ad) to do that)
- As to why (I think) that's necessary, there are some pages in which the data which is fetched from the backend needs to be different depending on whether the user is logged in or not.
- Though if there's a better approach, I would be interested in hearing about it

### Change Form

- [ ] The change has tests

# Related issue

- Resolve #140

### DIFF
--- a/client/src/stores/user.ts
+++ b/client/src/stores/user.ts
@@ -1,6 +1,6 @@
 import { defineStore } from 'pinia'
 import { useStorage } from '@vueuse/core'
-import { computed, ref } from 'vue'
+import { computed } from 'vue'
 import type { User } from '@/types/user'
 import server from '@/utils/server'
 import { isAxiosError } from 'axios'
@@ -10,7 +10,6 @@ export const useUserStore = defineStore('user', () => {
   const userData = useStorage<User | null>('userData', null)
   const accessToken = useStorage<string | null>('accessToken', null)
   const refreshToken = useStorage<string | null>('refreshToken', null)
-  const forceReload = ref<number>(0)
 
   const isLoggedIn = computed(() => userData.value !== null)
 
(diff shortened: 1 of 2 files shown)


cjnbennett:
Maybe I'm missing something, but couldn't you just reference `userData` (which should be reactive - so that any component which references it e.g. via `:key` will re-render when its value is updated on login/logout) rather than introducing a new variable `forceReload`?

HuxleyBerry:
> Maybe I'm missing something, but couldn't you just reference `userData` (which should be reactive - so that any component which references it e.g. via `:key` will re-render when its value is updated on login/logout) rather than introducing a new variable `forceReload`?

I tried that but couldn't get it to work since the ``User`` type seems to be incompatible with the types allowed for the ``key`` prop.

cjnbennett:
Fair. Maybe a variable storing the state of whether the user is logged in or not might make more sense than introducing a forceReload var? Otherwise it looks fine to me

HuxleyBerry:
> Fair. Maybe a variable storing the state of whether the user is logged in or not might make more sense than introducing a forceReload var? Otherwise it looks fine to me

Good idea, we already had a variable which stores whether the user is logged in. I had to convert it into a number in ``App.vue`` though since the ``key`` prop doesn't accept booleans.

cjnbennett (approved):
lgtm

client/src/stores/user.ts
```diff
@@ -17,6 +18,9 @@ export const useUserStore = defineStore('user', () => {
     userData.value = null
     accessToken.value = null
     refreshToken.value = null
+    if (router.currentRoute.value.meta.requiresAuth === true) {
```
jsun1590, line 21: `=== true` is unnecessary.

PR #153, opened 2025-02-01
Issue 94 api view to search for users

### LINKED ISSUE #94: API view to search for users
Create an API view which searches for the users whose names start with a given string. These users should be returned with the following information:
- Avatar
- Username
- User id
- Friend status. This represents one of four possibilities
  - The logged in user is friends with the given user
  - The logged in user has sent a friend request to the given user, but it is yet to be accepted.
  - The given user has sent a friend request to the logged in user, but it is yet to be accepted (also return the friendship id in this case).
  - No friend request has been sent from the logged in user to the given user or vice versa

The list of users should be truncated so that no more than 15 users are returned. However, make sure that a user whose name exactly matches the supplied string is not among those who is omitted due to the truncation.

**Edit**: superusers and users with ``is_active=False`` should not be included in this search

someonebear:
## Change Summary
Added view to search for users based on their username and a given search string.

### Change Form

- [X] The change has tests

## Other Information

# Related issue

- Resolve #94


server/bingo/utils.py
```diff
@@ -60,3 +60,29 @@ def check_line(indices):
...
+            Friendship, requester=current_user, receiver=user['user_id'])
+        if friendship:
+            if friendship.status == friendship.PENDING:
+                list_out.append((user, 'You have requested friendship.'))
+            elif friendship.status == friendship.ACCEPTED:
+                list_out.append((user, 'You are friends.'))
+        elif (friendship := get_or_none(Friendship, requester=user['user_id'], receiver=current_user)) is not None:
+            if friendship:
+                list_out.append(
+                    (user, f'Pending friendship request. {friendship.id}'))
```
HuxleyBerry: Can the message and id be separated in some more meaningful way so that the frontend doesn't have to extract stuff from the string?

server/bingo/views.py
```diff
@@ -321,3 +322,28 @@ def complete_challenge(request):
...
+    or 'Pending friendship request.'
+    """
+    USERS_RETURNED = 15
+    try:
+        query_string = request.data['query_string']
+    except KeyError:
+        return Response({'error': 'Field "query_string" is required in this request'},
+                        status=status.HTTP_400_BAD_REQUEST)
+    user_set = User.objects.filter(
+        username__istartswith=query_string, is_active=True, is_superuser=False)[:USERS_RETURNED]
```
HuxleyBerry: Is this guaranteed to get a user whose username exactly matches the query_string?

HuxleyBerry (changes requested):
Looks good. I'm just reiterating my requests I made in person in case you forget :)

server/bingo/utils.py
someonebear: Changed this. :) The response is also now dicts instead of tuples, so should be easier for front end crew.

server/bingo/views.py
someonebear: I've added an order by clause to the filter. Any exact matches should be at the top of the query set. Additionally the User model is ordered by points and then username, so I believe this will work. If you still want me to explicitly added the exact match in though, I can.

HuxleyBerry (approved):
Nice, LGTM!

PR #156, opened 2025-02-01
Issue 118 create api view to fetch other user s completed challenges

### LINKED ISSUE #118: Create API view to fetch other user's completed challenges
The view should take the username of a user as input and do the following things:
- Return the users first name, last name, avatar, bio, and number of points
- Calculate whether the logged in user (call them Y) has permission to view the completed challenges of this user (call them X)
  - If X has their visibility set to "BlueCrew Only", then Y can view X's completed challenges if and only if Y is staff
  - If X has their visibility set to "Friends Only", then Y can view X's completed challenges if and only if Y is staff or X is friends with Y
  - If X has their visibility set to "Public", then Y can always view X's completed challenge
- If Y does indeed have permission to view X's completed challenges, then for each challenge completed by X (i.e. each instance of the ``TileInteraction`` model which is linked to X), the following information should be returned:
  - The name of the challenge
  - The description of the challenge
  - The type of the challenge (connect/understand/act)
  - How many points the challenge was worth
  - The image X uploaded for the challenge
  - The date X started the challenge
  - The date X finished the challenge

The trickiest thing here is probably sending the image files to the frontend. I'm not sure of the details, but it could end up being hard to send multiple images together with json data. If so, an alternate option is for this view to not return the images, and instead have a separate view which would return a single image which would be called when the user clicks a button.

CohenRafiq:
## Change Summary
Added view to retrieve profile page info

### Change Form

- [x]The pull request title has an issue number
- [x] The change has tests

## Other Information
images are sent through as location


server/challenge_images/test_image.png
HuxleyBerry, line 1: These image files don't seem to be necessary?

server/bingo/views.py
```diff
@@ -270,6 +272,35 @@ def get_bingo_grid(request):
     return Response(grid, status=status.HTTP_200_OK)
 
 
+@api_view(['GET'])
+def get_profile_page(request, username):
+    try:
+        target_user = User.objects.get(username=username)
+    except User.DoesNotExist:
+        return Response({"error": "User not found"}, status=status.HTTP_404_NOT_FOUND)
```
HuxleyBerry: I think you can simplify this by using ``get_object_or_404()`` instead: https://docs.djangoproject.com/en/5.1/topics/http/shortcuts/#get-object-or-404.

server/bingo/utils.py
```diff
@@ -1,8 +1,18 @@
...
 from django.conf import settings
 
 
+def check_access(current_user, target_user):
+    if current_user.is_staff or current_user == target_user:
+        return True  # Can view any
+    elif (current_user.is_authenticated and
+          (Friendship.objects.filter(requester=current_user, receiver=target_user).exists() or
+           Friendship.objects.filter(requester=target_user, receiver=current_user).exists())):
+        return target_user.visibility != 0  # Can view any except staff
```
HuxleyBerry: These comments could be worded more clearly. For this one, I think you mean something like "Can view target_user's profile unless their visibility setting is set to 'staff only'" correct?

server/bingo/tests/test_retrieve_profile.py
```diff
@@ -0,0 +1,157 @@
...
+    def _assert_challenge_data(self, challenges):
+        """Helper method to assert challenge data"""
+        self.assertEqual(len(challenges), 16)
+        challenge = challenges[0]
+        self.assertEqual(challenge['name'], 'Challenge 0')
+        self.assertEqual(challenge['description'], 'Description 0')
+        self.assertEqual(challenge['challenge_type'], 'act')
+        self.assertEqual(challenge['points'], 5)
+        self.assertEqual(challenge['image'], '/path/to/image0.png')
+        self.assertEqual(challenge['date_completed'], '2025-01-18T11:00:00Z')
```
HuxleyBerry: It would be nice to create this date string programmatically rather than hardcoding it.

server/bingo/views.py
```diff
@@ -270,6 +272,35 @@ def get_bingo_grid(request):
     return Response(grid, status=status.HTTP_200_OK)
 
 
+@api_view(['GET'])
+def get_profile_page(request, username):
+    try:
+        target_user = User.objects.get(username=username)
```
HuxleyBerry: Can you add the requirement that the target_user is also not staff? That way, we'll make sure that it's not possible to view the profiles of staff.

server/bingo/tests/test_retrieve_profile.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class ProfilePageViewTestCase(TestCase):
+    def setUp(self):
+        self.client = APIClient()
+
+        # Create users with different visibility settings
+        self.users = {
+            'public': self._create_test_user(
+                'public',
```
HuxleyBerry: For consistency with your use of keyword arguments, could you have ``username='public'`` here? Same goes for below.

server/bingo/views.py
HuxleyBerry: Oh, also filter out the users with ``is_active=False``.

server/bingo/tests/test_retrieve_profile.py
```diff
@@ -0,0 +1,157 @@
...
+                    reverse('get_profile_page', kwargs={
+                            'username': username})
+                )
+                self.assertEqual(response.status_code, status.HTTP_200_OK)
+                self._assert_basic_user_info(
+                    response.data, target_user)
+                if i > j or (i == 0 and j == 0):
+                    self._assert_challenge_data(response.data['challenges'])
+                else:
+                    self.assertEqual(response.data['challenges'], [])
```
HuxleyBerry: This is nice and concise for what it does, but it is a bit hard to understand. Could you add some comments which explain a bit more about what is going on?

HuxleyBerry (changes requested):
Great work, this was a tricky issue! I've left some comments about some minor stuff.

server/bingo/views.py
```diff
@@ -274,10 +274,7 @@ def get_bingo_grid(request):
 
 @api_view(['GET'])
 def get_profile_page(request, username):
-    try:
-        target_user = User.objects.get(username=username)
-    except User.DoesNotExist:
-        return Response({"error": "User not found"}, status=status.HTTP_404_NOT_FOUND)
+    target_user = get_object_or_404(User, username=username, is_active=True)
```
HuxleyBerry: Can you add that the user isn't staff to the query?

server/bingo/tests/test_retrieve_profile.py
```diff
@@ -122,20 +126,39 @@ def _create_super_user(self):
...
                 self._assert_basic_user_info(
                     response.data, target_user)
-                if i > j or (i == 0 and j == 0):
-                    self._assert_challenge_data(response.data['challenges'])
+                if role_idx > visibility_idx or (role_idx == 0 and visibility_idx == 0):
+                    self._assert_challenge_data(
+                        response.data['challenges'])  # Only if have access
                 else:
+                    # If no access
                     self.assertEqual(response.data['challenges'], [])
```
HuxleyBerry: Can you instead assert that the length of ``response.data['challenges']`` is 0?

server/bingo/tests/test_retrieve_profile.py
```diff
@@ -97,7 +98,10 @@ def _assert_challenge_data(self, challenges):
         self.assertEqual(challenge['challenge_type'], 'act')
         self.assertEqual(challenge['points'], 5)
         self.assertEqual(challenge['image'], '/path/to/image0.png')
-        self.assertEqual(challenge['date_completed'], '2025-01-18T11:00:00Z')
+        self.assertEqual(challenge['date_completed'],
+                         DateTimeField().to_representation(
+                             datetime(2025, 1, 18, 11, 0, tzinfo=timezone.utc))
```
HuxleyBerry: It would be better to avoid repeating this datetime object. Maybe you could instead define it in the ``setUp`` method?

HuxleyBerry (changes requested):
The explanatory comments are great! I've requested some more minor stuff.

server/bingo/views.py
CohenRafiq: Can you explain why the user can't be staff? Can you not look at the profile of staff members?

HuxleyBerry (changes requested):
You've committed a bunch of image files. Ready to merge other than that.

HuxleyBerry (approved):
LGTM

PR #168, opened 2025-02-05
Issue 89 Deployment: Setup the compose files for production environment

### LINKED ISSUE #89: Deployment: Setup the compose files for production environment
Create a docker compose file which runs the following services:
- The server (which should be pulled from the latest image in dockerhub as facilitated by #88 )
- The database
- Nginx to serve the django admin dashboard and the django api
- Watchtower for automatic refresh of deployment

We already have a compose file, so to distinguish this new one, it should be called ``docker-compose-prod.yml``.

It could be helpful to look at a past cfc project, but try to understand rather than just copy: https://github.com/codersforcauses/csf/blob/main/docker-compose-prod.yml

melbauerr:
## Change Summary
Created a new Docker compose file called `docker-compose-prod.yml` based on past cfc project. Includes:
- The server
- The database
- Nginx
- Watchtower

Other changes include:
- Updated postgres image to later version
- Changed penultimate pair of lines from port 8000 to 8081 in `docker/server/entrypoint.sh`

### Change Form
- [❌] The pull request title has an issue number
- [ ] The change has tests

## Other Information
- Deleted unecessary files

# Related issue

- Resolve #89


HuxleyBerry (approved):
LGTM

**Edit**: for future readers, I shouldn't have approved the PR at this point since the db container had its 5432 port exposed, which was both unnecessary and potentially a security risk. Fixed now though.

PR #173, opened 2025-02-06
Fixed birthdate not being affected by form.

### LINKED ISSUE #160: Birthdate incorrectly stored in database
The dob entered in the registration modal doesn't affect the birthdate that ends up getting stored in the database.

someonebear:
## Change Summary
Birthdate field in model will now be affected by form.

### Change Form

- [ ] The change has tests

## Other Information
Registration form on the website does not give useful error messages. It just says registration failed or something. Does not help the user understand what they missed. Additionally, there is no way to tell which fields are actually required on the form. Right now all that is required is username, email, and password, but the form does not communicate this well. 

Edit: Oops just saw that there is already a pull request fixing this. Please ignore :)

# Related issue

- Resolve #160

### DIFF
--- a/server/bingo/serializers.py
+++ b/server/bingo/serializers.py
@@ -15,6 +15,8 @@ class UserRegisterSerializer(serializers.ModelSerializer):
                   'password', 'indigenous_identity', 'gender_identity']
         extra_kwargs = {
             'password': {'write_only': True},
+            'first_name': {'required': True},
+            'last_name': {'required': True},
         }
 
     def create(self, validated_data):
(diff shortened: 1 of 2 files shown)


server/bingo/serializers.py
```diff
@@ -3,35 +3,41 @@
...
 class UserRegisterSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
-        fields = ['username', 'first_name', 'last_name', 'email',
+        fields = ['username', 'first_name', 'last_name', 'email', 'birthdate',
                   'password', 'indigenous_identity', 'gender_identity']
         extra_kwargs = {
             'password': {'write_only': True},
-            'first_name': {'required': True},
-            'last_name': {'required': True}
```
HuxleyBerry: I believe the client would like first and last names to be required when signing up, so you can keep these lines. Doesn't matter too much though since the enforcement can be done on the frontend.

Could make birthdate required too, but I need to check with the client whether that's what they want.

HuxleyBerry (approved):

server/bingo/serializers.py
someonebear: Done 👍 : )

PR #176, opened 2025-02-06
Issue 128 - Refactored views.py

### LINKED ISSUE #128: Refactor views file
The ``views.py`` file is already quite long and is only going to get longer. The goal of this issue is to split the ``views.py`` file into multiple files. We could group together views which are in the same "feature group". These feature groups could be something like:
- User management
- Friends
- Leaderboard
- Bingo and Challenges
- Email stuff

someonebear:
## Change Summary
Separated views.py into separate files and moved into views/
Fixed urls.py with new views dir.
Moved utils.py into views/

### Change Form

- [X] The change has tests

## Other Information

# Related issue

- Resolve #128


HuxleyBerry (approved):
Thanks, LGTM!

server/bingo/views/bingo_views.py
```diff
@@ -0,0 +1,121 @@
...
+        return Response(status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+    try:
+        TileInteraction.objects.create(
+            user=request.user, position=challenge_index, grid=grid)
+    except IntegrityError:
+        # Throw an error if there is already an interaction between that user and that challenge
+        return Response(status=status.HTTP_409_CONFLICT)
+
+    return Response(status=status.HTTP_200_OK)
```
Fil-Naj: We can consolidate the multiple try catch blocks into one, and catch multiple exceptions so that we can still return individual errors code. It helps clean up the method. How does this look:
```py
def start_challenge(request):
    try:
        challenge_index = int(request.data["position"])

        if challenge_index not in range(16):
            return Response(status=status.HTTP_422_UNPROCESSABLE_ENTITY)

        grid = BingoGrid.objects.get(is_active=True)

        TileInteraction.objects.create(
            user=request.user, position=challenge_index, grid=grid
        )

    except (ValueError, KeyError):
        return Response(status=status.HTTP_422_UNPROCESSABLE_ENTITY)
    except ObjectDoesNotExist:
        return Response(status=status.HTTP_500_INTERNAL_SERVER_ERROR)
    except IntegrityError:
        return Response(status=status.HTTP_409_CONFLICT)

    return Response(status=status.HTTP_200_OK)
```

Here, the actions we want to perform on the request are clear, easier to read, and contained within one code block instead of 3.

> KeyError was handled in case the key "position" was not in the `request.data`

Thoughts?

By the way, great use of the early returns!

server/bingo/views/bingo_views.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@api_view(['GET'])
+def get_bingo_grid(request):
+    """
+    This view returns a dictionary with two keys.
```
Fil-Naj: I appreciate the fact it looks and, well...is, a dictionary on this side in python, but what we are actually sending back is a JSON object (once the dictionary has been serialised into JSON). JSON works in a key-value pair structure like a dictionary. Each key-value pair is commonly referred to as a "field" in the JSON object. So, instead of saying "This returns a dictionary with two keys", we can say "returning data with the following fields"

server/bingo/views/user_views.py
```diff
@@ -0,0 +1,67 @@
...
+@api_view(['PUT'])
+@permission_classes((permissions.IsAuthenticated, ))
+def update_user_preferences(request):
+    serializer = UpdatePreferencesSerializer(
+        instance=request.user, data=request.data)
+    if serializer.is_valid():
+        serializer.save()
+        return Response(status=status.HTTP_200_OK)
+    else:
+        return Response(data=serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
Fil-Naj: nitpick but you dont need to add the `else`, similar to what you did above on line 17

Fil-Naj (commented):
Few minor comments but LGTM!

HuxleyBerry:
@Fil-Naj in case you're not already aware, the purpose of this issue was just to split up already existing code into multiple files. The snippets you reviewed come from multiple different people. But you do bring up some good (if minor) points, the fixing of which would be in the spirit of refactoring :). So @someonebear if you're happy to make the suggested changes even though that's not what you "signed up for", that would be great.

Fil-Naj:
@HuxleyBerry My bad...But code refactor is perfect time for code cleanup!

someonebear:
Yeah haha, I didn't write those views, but I've made the changes you've requested @Fil-Naj , lgtm. Nothing wrong with being thorough.

PR #178, opened 2025-02-10
Issue 76 make bingo page

### LINKED ISSUE #76: Make bingo page

mye9:
## Change Summary
Finish the implementation of both mobile and desktop version in blingo page

### Change Form

- [x] The change has tests

## Other Information

# Related issue

- Resolve #76


client/src/views/BlingoView.vue
```diff
@@ -1,59 +1,264 @@
...
+
+const selectedTile = ref<number | null>(null)
+const showChallengeCard = ref(false)
+const currentChallenge = ref({
+  title: 'Watch an Ocean Documentary',
+  points: 200,
+  type: 'understand' as const,
+  description:
+    'Here are some of our top picks! You can choose one of them or watch one of your own. Tell us what you thought and submit a picture.',
+  status: 'not started' as const,
```
HuxleyBerry: You can fix a type error in the ``handleStatusChange()`` function if you replace the two ``as const``s with ``'understand' | 'connect' | 'act'`` and ``'not started' | 'started' | 'completed'`` respectively. However, I believe it would be more elegant to avoid an ``as`` casting by instead defining an interface (maybe called ``ChallengeInfo``) and requiring the ref to hold values of that type. Also, there could be no challenge currently selected, so it would make more sense for this ref to have type something like ``ChallengeInfo | null``.

Btw, if you're repeating the types ``'understand' | 'connect' | 'act'`` and ``'not started' | 'started' | 'completed'``, you could consider defining types for those somewhere in the ``types`` directory and importing those types into this file so that you don't repeat yourself.

client/src/views/BlingoView.vue
```diff
@@ -1,59 +1,264 @@
...
+          <div class="header-section desktop-header">
+            <h1 class="blingo-title text-primaryBlue">Blingo</h1>
+            <h2 class="blingo-subtitle text-primaryGreen">Connecting to the ocean</h2>
+          </div>
+
+          <!-- Challenge Card -->
+          <div v-if="showChallengeCard" class="challenge-card-container desktop-challenge-card">
+            <ChallengeCard
+              v-bind="currentChallenge"
+              :is-logged-in="true"
```
HuxleyBerry: This shouldn't always be true. It should be true if and only if the user is logged in (i.e. ``userStore.userData`` isn't null).

client/src/views/BlingoView.vue
HuxleyBerry, line 1: I'm experiencing that the modal width changes when the challenge card component switches between stages. It should remain the same width throughout. 
Compare
image
and
image

HuxleyBerry, line 1: Minor thing, but when I select a bingo tile, it causes a slight position change in some of the other bingo tiles.

client/src/views/BlingoView.vue
```diff
@@ -1,59 +1,264 @@
...
+
+        <!-- Game Grid -->
+        <div class="game-grid desktop-game-grid">
+          <div class="grid-row" v-for="row in 4" :key="`row-${row}`">
+            <BlingoTile
+              v-for="(_, col) in 4"
+              :key="`tile-${row}-${col}`"
+              type="connect"
+              text="placeholder text"
+              status="not started"
```
HuxleyBerry: Instead of hardcoding this, can you add a ref which stores an array of 16 ``ChallengeInfo`` objects and have the props of this ``BlingoTile`` component bound to the ``ChallengeInfo`` object at the appropriate index?

HuxleyBerry (changes requested):

client/src/views/BlingoView.vue
HuxleyBerry, line 1: On small screen sizes, the bingo tiles don't all have the same size if a certain tile has a long title:

image

This should be fixed by making the text be cutoff if it overflows its allocated area

HuxleyBerry (changes requested):

client/vite.config.ts.timestamp-1739416248224-6a8a87005b11d.mjs
HuxleyBerry, line 1: This file shouldn't be committed.

client/src/components/ChallengeCard.vue
```diff
@@ -185,109 +203,72 @@ const finish = () => {
...
+  color: white !important;
+  min-width: 180px !important;
+  height: 50px !important;
+  font-size: 18px !important;
+  text-transform: none !important;
+  display: flex !important;
+  align-items: center !important;
+  justify-content: center !important;
+  letter-spacing: normal !important;
+  padding: 0 32px !important;
```
HuxleyBerry: If possible, get rid of these ``!important``s.

client/src/components/ChallengeCard.vue
```diff
@@ -90,16 +103,16 @@ const finish = () => {
       <div class="description">
         <v-card-text>{{ description }}</v-card-text>
       </div>
-      <v-card-actions>
-        <v-btn color="white" v-if="!isLoggedIn" @click="openLoginModal">Login</v-btn>
-        <v-btn color="white" v-else @click="startTask">Start</v-btn>
-      </v-card-actions>
+      <div class="button-container">
+        <v-btn v-if="!isLoggedIn" @click="openLoginModal" class="action-button">Login</v-btn>
+        <v-btn v-else @click="startTask" class="action-button">Start</v-btn>
```
HuxleyBerry: You should use the ``bg-color`` prop to set the colour of these buttons. At the moment, all the buttons are dark blue.

client/src/components/ChallengeCard.vue
```diff
@@ -185,109 +203,72 @@ const finish = () => {
...
+  justify-content: center;
+  align-items: center;
+  width: 100%;
+  margin: 20px 0;
+}
+
+.action-button {
+  font-family: 'Poppins', sans-serif !important;
+  border-radius: 50px !important;
+  background-color: rgb(var(--v-theme-primaryPink)) !important;
```
HuxleyBerry: Get rid of specifying ``background-color`` here since you will be giving the buttons their colours through the ``bg-color`` prop.

client/src/views/BlingoView.vue
HuxleyBerry, line 1: On mobile, it seems like the modal is slightly wider than the challenge card it contains. The result is that there's a shadow that sticks outside of the card.

image

HuxleyBerry (changes requested):
Nice work on the fixes! I've noticed a couple of minor additional problems.

HuxleyBerry (approved):
LGTM

Issue #97, opened 2025-01-11
Create main friends page


HuxleyBerry:
Prerequisite: #78 

The page should look like this on desktop:
![image](https://github.com/user-attachments/assets/f39105e1-72f9-40e5-a053-d5bef0295f79)

The mobile version of the page is split up into multiple "subpages" (see the figma). To facilitate this, you should create a ref which stores which of the subpages is currently open.

For the purposes of this issue, you don't need to make the search bar or the buttons function. This issue is all about laying out the pieces of the ui in the correct positions.

PR #184, opened 2025-02-11
Issue 157 connect friends page to backend

### LINKED ISSUE #157: Connect friends page to backend
#97 Needs to be done first. The backend should be involved in the following ways:
- The friends shown should be fetched from the backend
- Pressing the accept/reject/delete buttons should call the appropriate backend API endpoint
- Typing in the search bar should call the people-searching API endpoint

MahitGtg:
## Change Summary
- Connected backend with frontend for the friends page so now we can see our current friends, incoming requests and outgoing requests with all buttons working as intended
- Made a new serializer called `FriendshipUserSerializer() ` which stores friendship id for all the `friendshipComponent` buttons to work
- Refer #97 to see the functioning of all buttons

- ### Change Form

- [ ] The change has tests

## Other Information

# Related issue

- Resolve #157


client/src/views/FriendView.vue
HuxleyBerry, line 1: Can you add some text which says "No users found" if there do not match any users matching the search query (from the searchbar)?

client/src/views/FriendView.vue
```diff
@@ -1,94 +1,136 @@
...
+    : [],
+)
+
+const filteredSearchResults = computed(
+  () =>
+    searchResults.value?.filter(
+      (result) =>
+        !currentFriends.value.some((friend) => friend.userId === result.user_data.user_id),
+    ) ?? [],
+)
```
HuxleyBerry: I think you can avoid doing this computation on the frontend. The backend ``/user-search/`` endpoint should return the information about whether the user you have searched for is a friend.

However, I think there may be a bug in the ``check_friendship()`` function in ``utils.py``. If there is a friendship instance with you as the receiver and some other person as the requester, then that function gives the status ``'Pending friendship request.'`` when it should give ``'You are friends.'``.

client/src/views/FriendView.vue
```diff
@@ -1,94 +1,136 @@
...
-const handleDelete = (id: number) => {
-  console.log('Delete friend:', id)
+  try {
+    const response = await server.post<SearchResult[]>('/user-search/', {
+      query_string: searchQuery.value,
+    })
+    searchResults.value = response.data
+  } catch {
+    searchResults.value = []
+    error.value = 'Failed to search users'
```
HuxleyBerry: Instead of modifying the ``error`` ref, could you use the messageStore instead? The reason is that if the search fails, then the search bar disappears and just the error text is visible. The user is then unable to interact with the search bar -- the only action they can do is reload the page.

client/src/views/FriendView.vue
```diff
@@ -1,94 +1,136 @@
...
+        await server.post(`/request-friendship/${userId}/`)
+        break
+      case 'delete':
+      case 'dismiss':
+      case 'reject':
+        await server.delete(`/delete-friendship/${friendshipId}/`)
+        break
+    }
+    await fetchFriendsData()
+    if (action === 'send') await searchUsers()
```
HuxleyBerry: It would be good to avoid this extra fetching from the backend if possible. For example, if the user accepts a certain friend request, then once the post request to ``/accept-friendship/`` returns a success status code, you can move the corresponding ``FriendEntry`` object from the ``incomingRequests`` array to the ``currentFriends`` array.

If you do this, then the user will see the result immediately rather than having to wait for the friends to load again. That would have the affect of the friend components suddenly disappearing, so it would also be good to use the message component to show some sort of success message.

HuxleyBerry (changes requested):
Looks good! I have requested some changes.

client/src/views/FriendView.vue
```diff
@@ -1,94 +1,136 @@
...
+// Data fetching
+const fetchFriendsData = async () => {
+  isLoading.value = true
+  error.value = null
+
+  try {
+    const [friendsResponse, incomingResponse, outgoingResponse] = await Promise.all([
+      server.get<FriendEntry[]>('/friends/'),
+      server.get<FriendEntry[]>('/friends/requests/incoming/'),
+      server.get<FriendEntry[]>('/friends/requests/outgoing/'),
```
HuxleyBerry: Not your fault (and not really in the scope of this issue), but I've realised there's no point in having those three separate endpoints since they're always done together. So the backend should be modified to combined the three corresponding api views into one.

client/src/views/FriendView.vue
HuxleyBerry: I suppose the efficient way to do this would be for the ``handleFriendAction()`` function to also take the index of the friendship-related object in its containing array as a parameter. This index would be passed to the function by accessing it in the ``v-for``. Then, the friendship-related object could be removed using the ``splice()`` function.

client/src/views/FriendView.vue
```diff
@@ -1,94 +1,156 @@
...
+        if (arrayIndex !== undefined) {
+          const array = action === 'dismiss' ? outgoingRequests : incomingRequests
+          array.value.splice(arrayIndex, 1)
+        }
+        successMessage = action === 'dismiss' ? 'Request cancelled' : 'Request rejected'
+        break
+
+      case 'send':
+        await server.post(`/request-friendship/${userId}/`)
+        await searchUsers() // Still need to refresh search results
```
HuxleyBerry: Why do the search results need to be refreshed?

server/bingo/views/utils.py
```diff
@@ -96,19 +95,26 @@ def check_line(indices):
...
             elif friendship.status == friendship.ACCEPTED:
                 list_out.append(
                     {'user_data': user, 'status': 'You are friends.'})
+        # Check if current user is receiver
         elif (friendship := get_or_none(Friendship, requester=user['user_id'], receiver=current_user)) is not None:
-            if friendship:
+            if friendship.status == friendship.PENDING:
                 list_out.append(
                     {'user_data': user, 'status': 'Pending friendship request.', 'friendship_id': friendship.id})
+            elif friendship.status == friendship.ACCEPTED:  # Add this condition
```
HuxleyBerry: Get rid of the comment on this line.

client/src/views/FriendView.vue
```diff
@@ -1,94 +1,156 @@
...
+        successMessage = 'Friend request sent!'
+        break
+    }
+
+    // Show success message
+    messageStore.showMessage('Success', successMessage, 'success')
+  } catch {
+    // Remove the unused error parameter
```
HuxleyBerry: Unnecessary comment

client/src/views/FriendView.vue
```diff
@@ -1,94 +1,156 @@
...
 
-const handledefault = (id: number) => {
-  console.log('View default:', id)
+// Event handlers
+// First, add an index parameter to handleFriendAction
+const handleFriendAction = async (
+  action: FriendAction,
+  userId: number,
+  friendshipId?: number,
+  arrayIndex?: number, // New parameter
```
HuxleyBerry: Unnecessary comment

HuxleyBerry (changes requested):
Almost done, just pointed out some minor stuff.

HuxleyBerry (approved):
LGTM

PR #188, opened 2025-02-14
Issue 154 create profiles page

### LINKED ISSUE #154: Create profiles page
Create the page that allows users to view other users' profiles/completed challenges. This page should take a prop/query parameter which is the username of the user you are trying to view the profile of. If no query parameter is provided, the page should default to the currently logged in user. Use this method to gain access to the query params: https://router.vuejs.org/guide/essentials/passing-props#Function-mode.

Image

Use the ``ChallengeInteraction`` component made by Sam and Yuweng.

This page does not need to be connected to the backend yet, that can be done in a later issue. So for the time being, just use some placeholder data.

aradavi63:
## Change Summary
- Created a profile page
- Can be accessed through http://localhost:3000/profile/{username}
- Created a mock user database with the users 'johndoe' and 'marinebiologist'

### Change Form

- [ ] The change has tests

## Other Information
- I just decided to push this early to look for feedback, apparently Zichen still has some UI changes he wants to do but hasn't done yet
- If you log in as a user, navigating to http://localhost:3000/profile displays the profile page for a user called 'currentuser', and overrides your account, instead of actually displaying the profile of the logged in user, so this needs to be fixed

# Related issue

- Resolve #154


HuxleyBerry (approved):
LGTM. I'd like to get this page into main ASAP so that it can be connected to the backend. We can worry about the minor UI improvements later.

PR #202, opened 2025-02-15
Issue 119 facilitate email verification and password reset on the backend

### LINKED ISSUE #119: Facilitate email verification and password reset on the backend
Here's a non-exhaustive list of things to be done:
- Create a ``django.contrib.auth.tokens.PasswordResetTokenGenerator`` instance
- Modify the user registration view so that users are not marked as active
- Write a view which takes an email address and sends an email containing a password verification link to that address, provided that the user corresponding to that address exists and isn't already active
- Write a view which takes a user id and a reset token, checks if that token is valid for the user, and if so marks the user as active.
- Write a view which takes an email address and sends an email containing a password reset link to that address, provided that the user corresponding to that address exists
- Write a view which takes a user id, a reset token, and a new password, checks if that token is valid for the user, and if so assigns the user the new password.

See the django docs for information on sending emails: https://docs.djangoproject.com/en/5.1/topics/email/. For the time being, I suggest using the [console backend](https://docs.djangoproject.com/en/5.1/topics/email/#console-backend) for testing purposes.

SafetyInObscurity:
## Change Summary
Implemented email verification and password reset views

### Change Form

- [x] The change has tests

## Other Information
Environment variables will need to be updated from the example file, and the users model no longer has is_active defaulting to True

# Related issue

- Resolve #119


server/bingo/urls.py
```diff
@@ -1,35 +1,26 @@
...
     path('user-search/', user_views.find_user, name='user_search'),
-    path('update-bingo-grid/', bingo_views.update_bingo_grid,
-         name='update-bingo-grid'),
-    path('friends/all/', friends_views.get_all_friends_data, name='all_friends_data'),
-
+    path('update-bingo-grid/', bingo_views.update_bingo_grid, name='update-bingo-grid'),
+    path("email-validation", email_views.request_email_verification, name='request_verification'),
+    path("activate", email_views.confirm_email, name="confirm_email"),
+    path("request-reset", email_views.request_password_reset, name="request_password_reset"),
+    path("reset-passoword", email_views.reset_password, name="reset_password"),
```
HuxleyBerry: Can you add trailing slashes at the end of these urls for consistency with the rest?

HuxleyBerry: Also, there's a typo: "reset-passoword"

server/bingo/views/email_views.py
```diff
@@ -0,0 +1,128 @@
...
+    url = request.build_absolute_uri(f"{reverse("confirm_email")}?uid64={encoded_user}&token={token}")
+    content = render_to_string(
+        "verification.html",
+        context={"url": url}
+    )
+    message = EmailMessage(
+        "Bingo Email Verification",
+        content,
+        settings.ACCOUNTS_EMAIL,
+        list(email)
```
HuxleyBerry: This line should be ``[email]`` rather than ``list(email)`` since ``list(email)`` creates a list of all the characters in the email. You could also add a test which would have caught this error (checking if the sent email is correct).

server/bingo/views/email_views.py
```diff
@@ -0,0 +1,128 @@
...
+    url = request.build_absolute_uri(f"{settings.FRONTEND_URL}/{"reset_path"}/?uid64={encoded_user}&token={token}")
+    content = render_to_string(
+        "password_reset.html",
+        context={"url": url}
+    )
+    message = EmailMessage(
+        "Bingo Password Reset",
+        content,
+        settings.ACCOUNTS_EMAIL,
+        list(email)
```
HuxleyBerry: Same thing about ``[email]``.

server/bingo/views/email_views.py
HuxleyBerry, line 1: As a general comment, could you include a descriptive message in more of the responses (particularly the ones with status code 400 and 500)? That is important so that the frontend knows what to explain to the user.

server/bingo/views/email_views.py
```diff
@@ -0,0 +1,128 @@
...
+        token = request.data["token"]
+        password = request.data["password"]
+        uid = force_str(urlsafe_base64_decode(uid64))
+    except (KeyError, ValueError):
+        return Response(status=status.HTTP_400_BAD_REQUEST)
+
+    try:
+        validate_password(password)
+    except ValidationError:
+        return Response(status=status.HTTP_400_BAD_REQUEST)
```
HuxleyBerry: Can you include the validation error in the response here? That way the frontend will know what error message to show.

HuxleyBerry: This could be tidied slightly by combining the two try-except blocks into one try block with two except blocks.

HuxleyBerry (changes requested):
Nice work! I haven't finished going over everything yet, but before I go to bed I'll point out the things I've noticed so far.

server/bingo/views/email_views.py
```diff
@@ -0,0 +1,128 @@
...
+
+    user = get_object_or_404(User, user_id=uid)
+    if user.is_active:
+        return redirect(success_path)
+    if not email_verification_token_generator.check_token(user, token):
+        return Response(status=status.HTTP_404_NOT_FOUND)
+    user.is_active = True
+    user.save()
+    return redirect(success_path)
```
HuxleyBerry: Instead of returning a redirect on success, this view should return a response with a 200 status code.

server/bingo/views/email_views.py
```diff
@@ -0,0 +1,128 @@
...
+        message.send(fail_silently=False)
+    except SMTPSenderRefused:
+        return Response(status=status.HTTP_400_BAD_REQUEST)
+    except SMTPException:
+        return Response(status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+    return Response(status=status.HTTP_200_OK)
+
+
+@api_view(["GET"])
```
HuxleyBerry: This view should accept POST (or maybe PATCH) request rather than GET requests since it leads to a change of state in the server.

server/bingo/views/email_views.py
```diff
@@ -0,0 +1,128 @@
...
+    try:
+        email = request.POST['email']
+        user = User.objects.get(email=email)
+    except (ObjectDoesNotExist, KeyError):
+        return Response(status=status.HTTP_400_BAD_REQUEST)
+
+    token = default_token_generator.make_token(user)
+    encoded_user = urlsafe_base64_encode(force_bytes(user.pk))
+    # TODO Replace with link to the actual reset form
+    url = request.build_absolute_uri(f"{settings.FRONTEND_URL}/{"reset_path"}/?uid64={encoded_user}&token={token}")
```
HuxleyBerry: To match with what Sam and Yuweng have done on the frontend, the url should be ``f"{settings.FRONTEND_URL}/forgot-password?uid64={encoded_user}&token={token}"``

server/bingo/views/email_views.py
```diff
@@ -0,0 +1,128 @@
...
+        email = request.POST['email']
+        user = User.objects.get(email=email)
+    except (ObjectDoesNotExist, KeyError):
+        return Response(status=status.HTTP_400_BAD_REQUEST)
+
+    if user.is_active:
+        return Response(status=status.HTTP_400_BAD_REQUEST)
+    encoded_user = urlsafe_base64_encode(force_bytes(user.pk))
+    token = email_verification_token_generator.make_token(user)
+    url = request.build_absolute_uri(f"{reverse("confirm_email")}?uid64={encoded_user}&token={token}")
```
HuxleyBerry: To match with what Sam and Yuweng have done on the frontend, the url should be ``f"{settings.FRONTEND_URL}/verify-email?uid64={encoded_user}&token={token}"``

server/bingo/views/email_views.py
```diff
@@ -0,0 +1,128 @@
...
+from django.contrib.auth.tokens import default_token_generator
+from django.contrib.auth.password_validation import validate_password
+from django.urls import reverse
+from smtplib import SMTPException, SMTPSenderRefused
+
+
+@api_view(['POST'])
+def request_email_verification(request):
+    try:
+        email = request.POST['email']
```
HuxleyBerry: Use ``request.data`` rather than ``request.POST``. The same goes for the other views.

server/bingo/models.py
```diff
@@ -90,7 +90,7 @@ class IndigenousIdentity(models.IntegerChoices):
     avatar = models.IntegerField(default=0, choices=map(
         lambda i: (i, f"Avatar {i}"), range(6)))
 
-    is_active = models.BooleanField(default=True)
+    is_active = models.BooleanField(default=False)
```
HuxleyBerry: As we already discussed, it would be better leaving the model as it is and instead modifying the UserRegisterSerializer to make users automatically inactive upon registering.

HuxleyBerry (changes requested):

server/bingo/views/email_views.py
SafetyInObscurity: Trouble is, this is going to be directly accessed by users clicking the link. Their browsers will send a GET request for the page at the address.

server/bingo/views/email_views.py
SafetyInObscurity: Again, this is a page that users are currently directly accessing via browser. They'd see a 200 response as an error.

server/bingo/views/email_views.py
HuxleyBerry: Maybe I wasn't clear about this earlier, but we are using the frontend as an intermediary in the email verification and forgot password processes. The link you provide in your email template sends the user to a page for email verification in the frontend. This frontend page then sends a request to the backend using the query params in its url as data. That's the request I'm asking to be a POST request.

Why use the frontend as an intermediary? Partly because we've already started doing it that way in other issues, but the main reason is that the user never has to visit the backend. It could be off putting to the user to click on a link without the ``blingo.com.au`` domain they were expecting.

SafetyInObscurity: Ok, I'll rework it. Do you know the path they're using for the email verification?

HuxleyBerry: I think it'll be ``f"{settings.FRONTEND_URL}/verify-email?uid64={encoded_user}&token={token}"``

server/bingo/urls.py
SafetyInObscurity: Oops

HuxleyBerry (approved):
Nice, LGTM!

PR #206, opened 2025-02-16
Friends page enhancement

HuxleyBerry:
## Change Summary
- Made it so that when the user interacts with a search result (e.g. presses the "accept" button on a search result), the search results immediately update (done without an additional request to the backend)
- However, the effect of pressing such a button should be that the friends list changes when the user closes the search. As far as I can tell, there's no efficient and scalable way of doing this only on the frontend since (In the "accept" case) you don't know the index of the person within the ``incomingFriends`` array. Thus, the friends are fetched from the backend again.
- Added a spinner which appears while the search results are still loading

### Change Form

- [ ] The change has tests

## Other Information


MahitGtg:
LGTM!

PR #209, opened 2025-02-18
Issue 112 connect bingo page to backend

### LINKED ISSUE #112: Connect bingo page to backend
## Prerequisites

#76 #56 #62 #75 

## Issue
Connect the bingo page frontend to the backend. There are three parts to this:
- When the page loads, the page should call the ``api/bingo-grid/`` endpoint to fetch the current bingo tiles. The page should then render the appropriate tiles.
- If the user presses the start button, the ``api/start-challenge/`` endpoint should be called.
- If the user presses the complete button, the the ``api/complete-challenge/`` endpoint should be called. This is trickier since this will involve sending a file to the backend. Some information on achieving this can be found [here](https://stackoverflow.com/a/51109645) and [here](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects).

Also, if the backend responds with a success message after the user starts or completes a challenge, the rendered bingo grid should update appropriately.

someonebear:
## Change Summary
Bingo page is now connected to back end.

### Change Form

- [ ] The change has tests

## Other Information
Consent form field does not exist on challenge card component. Feedback field does not exist in backend models? Not sure if we discussed this on saturday. I forgot which ones you said already had issues so if you could create these new issues that would be great :))))

# Related issue

- Resolve #112


HuxleyBerry:
> Consent form field does not exist on challenge card component. Feedback field does not exist in backend models? Not sure if we discussed this on saturday. I forgot which ones you said already had issues so if you could create these new issues that would be great :))))

A feedback field was added into main in a recent pull request (the field is named ``description``). You're right about the consent checkbox being missing though.

someonebear:
Yep I see now. The feedback is properly stored in the backend now.

client/src/views/BlingoView.vue
```diff
@@ -165,13 +76,14 @@ const handleStatusChange = (newStatus: 'not started' | 'started' | 'completed')
               <!-- Header Section -->
               <div class="header-section desktop-header">
                 <h1 class="blingo-title text-primaryBlue">Blingo</h1>
-                <h2 class="blingo-subtitle text-primaryGreen">Connecting to the ocean</h2>
+                <h2 class="blingo-subtitle text-primaryGreen" h>Connecting to the ocean</h2>
```
HuxleyBerry: The ``h`` should be removed.

client/src/components/ChallengeCard.vue
```diff
@@ -48,6 +48,7 @@ const props = defineProps<{
   description: string
   status: ChallengeStatus
   isLoggedIn: boolean
+  position: number | null
```
HuxleyBerry: IMO it would be better to avoiding allowing ``position`` to be null. I suggest removing the ``| null`` and making the necessary change in ``BlingoView.vue`` (Adding an extra condition to the ``v-if`` surrounding the ChallengeCard component)

client/src/views/BlingoView.vue
```diff
@@ -1,133 +1,40 @@
...
-  },
-])
+const challengeInfos = ref<ChallengeInfo[]>([])
+const messageStore = useMessageStore()
+
+const fetchBingoGrid = () => {
+  server
+    .get('/bingo-grid/')
+    .then((res) => {
+      res.data.challenges.forEach((challenge: ChallengeInfoAPI) => {
```
HuxleyBerry: Using ``map()`` instead of ``forEach()`` would be more elegant

client/src/views/BlingoView.vue
```diff
@@ -165,13 +76,14 @@ const handleStatusChange = (newStatus: 'not started' | 'started' | 'completed')
...
                 <h1 class="blingo-title text-primaryBlue">Blingo</h1>
-                <h2 class="blingo-subtitle text-primaryGreen">Connecting to the ocean</h2>
+                <h2 class="blingo-subtitle text-primaryGreen" h>Connecting to the ocean</h2>
               </div>
 
               <!-- Challenge Card -->
               <div v-if="showChallengeCard" class="challenge-card-container desktop-challenge-card">
                 <ChallengeCard
                   v-bind="currentChallenge"
+                  v-bind:position="selectedTile"
```
HuxleyBerry: Definitely a nit-pick, but could you use the abbreviated form of ``v-bind``?

client/src/views/BlingoView.vue
```diff
@@ -245,6 +157,7 @@ const handleStatusChange = (newStatus: 'not started' | 'started' | 'completed')
         <div v-if="showChallengeCard" class="challenge-card-overlay">
           <ChallengeCard
             v-bind="currentChallenge"
+            v-bind:position="selectedTile"
```
HuxleyBerry: Same

client/src/views/BlingoView.vue
HuxleyBerry, line 1: On page load, I get a bunch of warnings in the console like ``[Vue warn]: Missing required prop: "type"``. These warnings are from the fact that the ``BingoTile`` components have their ``v-bind`` set to elements of the array ``challengeInfos`` which is initially empty before the data gets fetched. One way to fix this would be to not show the bingo tiles until the data gets fetched, and show a loading spinner in the meantime. You can copy how to do this from the leaderboard page.

client/src/components/ChallengeCard.vue
HuxleyBerry, line 1: The buttons in this component should have loading spinners while the frontend waits for a response from the backend.

HuxleyBerry (changes requested):
Works well for me! I've requested some small improvements.

someonebear:
Made all changes except for the last one. I'm not too sure how to do that, and I'm not sure I can in time for your deadlines. :)

HuxleyBerry (approved):
LGTM!

PR #221, opened 2025-02-22
Issue 92 animate bingos and completion of challenges

### LINKED ISSUE #92: Animate bingos and completion of challenges
When the user completes a challenge, it would be nice to have an animation which shows the user being rewarded points. Also, when the user achieves a bingo, it would nice for the row/column/diagonal to be "highlighted" in some way and the number of bonus points received to be shown.

mye9:
## Change Summary

### Change Form

- [ ] The change has tests

## Other Information
Animation Done with highlighted tiles when bingo is done

# Related issue

- Resolve #92


client/src/views/BlingoView.vue
HuxleyBerry, line 1: I had something a little different in mind regarding the highlighting of completed bingos. That's probably my fault for not writing a more detailed issue description. Anyway, what I had in mind is that a row/column/diagonal in which a bingo is achieved is highlighted *just* at the moment the bingo is achieved. So if you were to navigate to a different page and then return to the blingo page, no rows/columns/diagonals would be highlighted.

It looks like at the moment you have some code to figure out if there are currently any rows/columns/diagonals with all the tiles completed. I think you can get rid of that code and just use the information returned by the ``complete-challenge`` endpoint. See [here](https://github.com/codersforcauses/bluecrew/blob/52030a0109d68d46086bece471453f0eba703f58/server/bingo/views/bingo_views.py#L74) for a description of the data that that endpoint returns. So instead of calculating on the frontend if any bingos have been achieved, you simply see if the ``complete-challenge`` endpoint reports any bingos, and if so, you trigger a temporary animation which involves making some of the bingo tiles highlighted.

In the event multiple bingos are achieved simultaneously, I think it would it look best if the highlighting animations happened one-by-one for each row/column/diagonal a bingo was achieved in. You might as well fix the order of these animations as the row bingo, then the column bingo, then the diagonal bingo, and then the full bingo (where of course you only do each step if such a bingo is actually achieved).

Since the row/column/diagonal is temporary, we can get away with a more intrusive animation. The current highlighting is pretty subtle.

client/src/components/BingoTile.vue
HuxleyBerry, line 1: I like the exploding animation! My only concern is that at the start of the animation temporarily causes the row the tile is in to shrink inwards. The animation would look cleaner if the grid was able to maintain its shape throughout.

HuxleyBerry (changes requested):
Thanks for working on this issue Ming!

client/src/views/BlingoView.vue
```diff
@@ -38,6 +39,69 @@ const fetchBingoGrid = async () => {
     })
 }
 
+const bingoRows = ref<boolean[]>(Array(gridSize).fill(false))
+const bingoCols = ref<boolean[]>(Array(gridSize).fill(false))
+const bingoDiagonal = ref<boolean[]>([false, false]) // [↘, ↙]
+
+const highlightBingo = async () => {
+  try {
+    const response = await server.patch('/complete-challenge/')
```
HuxleyBerry: This component should not call the ``/complete-challenge/`` endpoint. The ``ChallengeCard`` component already calls that endpoint, so you should receive an emit from that component with the bingo information instead.

client/src/views/BlingoView.vue
HuxleyBerry, line 1: The exploding animation (on completion of a challenge) does not occur at the tile position. Instead, I consistently see the expansion animation happening somewhere off to the right of the grid.

HuxleyBerry (changes requested):

HuxleyBerry (approved):

PR #226, opened 2025-02-23
Issue 131 create scheduled task to delete unverified users

### LINKED ISSUE #131: Create scheduled task to delete unverified users
Once #119 is done, we will have the problem that a user who fails to verify their email after registering will be stored in the db forever. Not only does this take up extra space, but more significantly, no user will be able to sign up using that email in the future. Due to that, it would be possible to "lock people out" from ever being able to sign up with a given email. A solution to this problem is to have a task which runs regularly (maybe once per day) that deletes the users who have ``is_active`` set to False from the db.

This usage of ``is_active`` isn't quite in line with the recommendations of the [django docs](https://docs.djangoproject.com/en/5.1/ref/contrib/auth/#django.contrib.auth.models.User.is_active), but I think it'll be fine for our purposes.

HuxleyBerry:
## Change Summary
- Created a task which deletes users who have been inactive in the db for more than a day
- Wrote a test for that task
- Created a management command which starts a schedule to run that task once per day
  - this command is run automatically in prod

### Change Form

- [x] The change has tests

## Other Information

# Related issue

- Resolve #131


server/bingo/management/commands/schedule_tasks.py
```diff
@@ -0,0 +1,17 @@
+from django.core.management.base import BaseCommand
+from django_q.models import Schedule
+
+
+class Command(BaseCommand):
+    help = "Schedule the task of deleting inactive users who have existed in the database for more than a day."
+
+    def handle(self, *args, **options):
+        if not Schedule.objects.filter(name="Start Matchmaking").exists():
+            print("Schedule does not exist, making now")
```
junhyap: You could use `self.stdout.write()` for consistent Django management output but its a bit pedantic.

server/bingo/management/commands/schedule_tasks.py
```diff
@@ -0,0 +1,17 @@
...
+            Schedule.objects.create(
+                func='bingo.tasks.delete_inactive',
+                name="Delete inactive users",
+                schedule_type=Schedule.DAILY,
```
junhyap, line 16: `Schedule.DAILY` creates a task that runs every 24 hours, but it’s good practice to set `next_run` explicitly.

junhyap: @frinzekt might have some better thoughts on this tbh.

server/bingo/management/commands/schedule_tasks.py
```diff
@@ -0,0 +1,17 @@
+from django.core.management.base import BaseCommand
+from django_q.models import Schedule
+
+
+class Command(BaseCommand):
+    help = "Schedule the task of deleting inactive users who have existed in the database for more than a day."
+
+    def handle(self, *args, **options):
+        if not Schedule.objects.filter(name="Start Matchmaking").exists():
```
junhyap: Should this be `Start Matchmaking`? From my understanding of what you're trying to do, maybe `"Delete inactive users"` would be a better suited name. Correct me if I'm wrong tho.

server/bingo/migrations/0020_user_created_at.py
```diff
@@ -0,0 +1,22 @@
...
+    dependencies = [
+        ("bingo", "0019_alter_tileinteraction_image"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="created_at",
+            field=models.DateTimeField(
+                auto_now_add=True, default=django.utils.timezone.now
```
junhyap, line 18: `auto_now_add=True` automatically sets the field to the current timestamp when a record is created. The `default=django.utils.timezone.now` argument is redundant and unnecessary.

Django doesn't allow `default` and `auto_now_add=True` together in newer versions but I suspect it's not a big deal here?

server/bingo/tasks.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def delete_inactive():
+    to_delete = User.objects.filter(is_active=False, created_at__lte=(
+        timezone.now() - timedelta(days=1)))
+    if count := to_delete.count():
+        to_delete.delete()
+        print(f"{count} inactive users deleted.")
+    else:
+        print("There are currently no inactive users.")
```
junhyap: Again, `print()` is fine for development but logs _should_ be used in production as this can make debugging easier when running the task via a scheduler

server/bingo/tasks.py
```diff
@@ -0,0 +1,13 @@
...
+from datetime import timedelta
+from django.utils import timezone
+
+
+def delete_inactive():
+    to_delete = User.objects.filter(is_active=False, created_at__lte=(
+        timezone.now() - timedelta(days=1)))
+    if count := to_delete.count():
+        to_delete.delete()
+        print(f"{count} inactive users deleted.")
```
junhyap: Calling `.count()` actually fetches all matching records, which is inefficient. Using `.exists()` first prevents an unnecessary database query.

So `.delete()` actually returns a tuple of (number of records deleted, details). Therefore instead of fetching .count() first, we can capture this return value.

Maybe you could write it something like

```
if to_delete.exists():
        deleted_count, _ = to_delete.delete()
```

server/bingo/tests/test_inactive_user_deletion.py
```diff
@@ -0,0 +1,32 @@
...
+        self.new_inactive_user = User.objects.create(
+            username="user3", password="verysecure123", email="email3@example.com", is_active=False)
+        with freeze_time(timezone.now() - timedelta(days=2)):
+            self.old_active_user = User.objects.create(
+                username="user4", password="verysecure123", email="email4@example.com")
+        with freeze_time(timezone.now() - timedelta(hours=1)):
+            self.recent_active_user = User.objects.create(
+                username="user5", password="verysecure123", email="email5@example.com", is_active=False)
+        self.new_active_user = User.objects.create(
+            username="user6", password="verysecure123", email="email6@example.com", is_active=False)
```
junhyap: Not sure if this is an oversight or intended but `recent_active_user` and `new_active_user` are created as inactive (`is_active=False`). 

Active users should be `is_active=True` right?

server/bingo/tests/test_inactive_user_deletion.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def test_deletion_result(self):
+        delete_inactive()
+        # The only user who should be deleted is the user created more than one day ago who is inactive
+        self.assertEqual(set(u.username for u in User.objects.all()), {
```
junhyap: Using `set()` assumes uniqueness, but if there's a bug where multiple users are deleted incorrectly, the test might still pass.

Might be a better idea to check deletions explicitly.

junhyap: You could also just check that only the old inactive user was deleted, instead of checking all users left in the database.

Something like this maybe:
```
    def test_deletion_result(self):
        delete_inactive()

        # Ensure the old inactive user is deleted
        self.assertFalse(User.objects.filter(username="user1").exists(), "Old inactive user was not deleted.")

        # Ensure all other users are still in the database
        remaining_users = set(User.objects.values_list("username", flat=True))
        expected_users = {"user2", "user3", "user4", "user5", "user6"}

        self.assertEqual(remaining_users, expected_users, "Unexpected users were deleted.")
```

junhyap (commented):
Some minor comments but overall good! Great job my goat.

I'm also not a Django expert so might be nice if @frinzekt could weigh in on this.

server/bingo/management/commands/schedule_tasks.py
HuxleyBerry: Yep you're right, was a copy paste mistake

server/bingo/migrations/0020_user_created_at.py
HuxleyBerry: This is a migration file so I don't have much control of what is in it. Anyway, I think it makes sense for a default to be present here, since this migration adds a non-nullable column to the user table. The pre-existing rows need some value put in their ``created_at`` column.

junhyap: oh right duh ok

junhyap (approved):

PR #242, opened 2025-03-06
Issue 205 update banners

lumenesce:
## Change Summary
- changed banners so each page has a unique banner (7 total)
- included banners in mobile view for profile and preference pages
- got rid of whitespace above banners in mobile view!
- tried to improve consistency related to padding, removed second img element
- removed unused classes
- changed "Go to bingo" to "Go to blingo" on the landing page

### Change Form

- [ ] The change has tests

## Other Information


client/src/App.vue
```diff
@@ -24,9 +24,3 @@ const userStore = useUserStore()
   <RegisterModal v-if="modalStore.currentModal === 'register'" @close="modalStore.closeModal" />
   <LoginModal v-if="modalStore.currentModal === 'login'" @close="modalStore.closeModal" />
 </template>
-
-<style scoped>
-.main-content {
-  margin-top: 56px;
-}
-</style>
```
HuxleyBerry, line 32: I don't think you should have deleted this. The result is that the page's content starts at the top of the screen which is behind the navbar. That's fine on mobile, but not good on desktop. For example, the top part of the banner is hidden behind the navbar on the desktop landing page. I suggest adding this style back, potentially with it being conditionally present if the screen isn't mobile sized.

It looks like you made some changes in other files which involved shifting the content down to compensate for this change, so those might need to be reverted too.

HuxleyBerry (changes requested):

HuxleyBerry (approved):
LGTM!

